PR #7, opened 2018-09-10
Building and testing using Go 1.11

asdine:


.travis.yml
```diff
@@ -16,7 +16,8 @@ install: dep ensure
 
 go:
   - 1.9
-  - "1.10"
+  - '1.10'
```
YassineE: 🌷 I think that in `yaml` you can skip the quote.

tealeg: Didn't there used to be a bug if you mentioned Go versions after 1.9 without quoting?

asdine: Actually no you have to, otherwise it will parse it as a number, remove the `0` at the end and use `go1.1`.

https://twitter.com/golang_news/status/967370924218634243

YassineE: I think the bug is related to the format of the file and it happens if the format is json.
For example, in [confita](https://github.com/heetch/confita/blob/master/.travis.yml#L10) we use the versions without quote and it works well.

YassineE: Oh i wasn't aware about this tweet.
So i think that it works on confita because we have the trailing `.x` ?

asdine: > So i think that it works on confita because we have the trailing .x ?

Exactly, and if you take a look at the previous version of the Confita travis config you can notice the quotes on the Go 1.10 version

https://github.com/heetch/confita/blob/c0289b95b1f484274f5e33ba0c2265a7b929002e/.travis.yml#L25

Issue #15, opened 2018-11-02
Define predicates for s-expression lexer


tealeg:
Define predicate functions to be used in a lexical scan of symbollic expressions.  Each of these should be able to identify a character type that matches one of the Token constants defined by #14.  Any character not matched by one of these must be part of a symbol or string.

Each predicate will accept a single rune as its input, and return a boolean. 

- `isWhitespace`
- `isNumber`
- `isString`
- `isBoolean`
- `isLParen`
- `isRParen`
- `isSymbol`
- `isComment`

Issue #14, opened 2018-11-02
Define lexical primitives for symbolic expressions


tealeg:
Define the fundamentals required to build an s-expression lexer:

- Token type
-  Token constants for: 
   -`(`, `)`  - bounding a list-form s-expression
  - '"` - bounding a string value
  - Whitespace chars
  - arabic digits, including negatives and floating point forms
 - `#t` and `#f` forms for booleans
 - `;` commenting the remainder of a line
 - Symbols

PR #18, opened 2018-11-02
S-Expression lexer primitives and predicates

tealeg:
This PR puts in place the first building block of the lexer - a series of Token types and predicates that identify the token about to be scanned.  

Note that we already see a potential "special case" in our language - the `-` character can start a number or be an symbol referencing an operator.  Lisp implimentations solve this issue in varied ways, most rely on knowing that function references look different (i.e Common Lisp would referencet the `-` function as `#'-`) and that implicit invocations only occur on symbols that appear at the head of a list-form s-expression.  Some go as far as doing in-line expansions in the reader to convert negated numbers into the logical form (i.e. converting `-1` to `(- 0 1)`).    We'll probably just force the token to the correct type in the Parser instead. 

Fixes: #14 
Fixes: #15


sexpr/lexer.go
```diff
@@ -0,0 +1,64 @@
...
+	NUMBER
+	BOOL
+	COMMENT
+	SYMBOL
+)
+
+// isWhitespace returns true if the rune is the first rune of a
+// Whitespace sequence.
+func isWhitespace(r rune) bool {
+	return r == ' ' || r == '\t' || r == '\r' || r == '\n'
```
asdine: :tulip: The unicode package provides a [`IsSpace` function](https://golang.org/pkg/unicode/#IsSpace) that deals with additional special characters. Do you think we should use it?

sexpr/lexer.go
```diff
@@ -0,0 +1,64 @@
+package sexpr
```
asdine: The `sexpr` package is directly related to the `rule` package, what about moving it under `rule/sexpr` ?

sexpr/lexer_internal_test.go
```diff
@@ -0,0 +1,146 @@
...
+	require.False(t, isLParen('\n'))
+	require.False(t, isLParen('-'))
+	require.False(t, isLParen('a'))
+	require.False(t, isLParen('"'))
+	require.False(t, isLParen(')'))
+	require.False(t, isLParen('_'))
+	require.False(t, isLParen('0'))
+	require.False(t, isLParen('#'))
+	require.False(t, isLParen(';'))
+}
```
YassineE: 🌷🌷 maybe add: `require.False(t, isLParen(')')`.

sexpr/lexer.go
tealeg: Yes!  This was buzzing in my head on Friday - I knew I'd seen it somewhere.

sexpr/lexer.go
tealeg: Good idea.

sexpr/lexer_internal_test.go
YassineE: My bad 🙈

Issue #16, opened 2018-11-02
Define Scanner and contiguous sequence readers for an symbolic expression lexer


tealeg:
Create a Scanner type for an s-expression Lexer.

The scanner will wrap a `bufio.Reader` and will support sequential character scanning, with backtracking capability.  Each rune will be read in turn by successive calls to `Scanner.Scan`, if a rune indicates the beginning of a contiguous sequence then we will call out to a function to gather all of that contiguous sequence, and return its literal content, along with the token type.

The contiguous sequence type scanners must be:
- `scanWhitespace`
- `scanComment`
- `scanDigit`
- `scanString`
- `scanSymbol`
- `scanBoolean`
Each will return a token type and a literal string containing the scanned text.

PR #29, opened 2018-11-09
Scanner and contiguous sequence readers for symbolic expression lexer 

tealeg:
This PR defines the scanner for the symbollic expression language to be deployed in the Regula UI.

There are essentially 3 layers to this:
  - A `Scanner` struct that wraps a `io.Reader` in a `bufio.Reader` and uses it to read the stream a rune at a time (supporting backtracking as needed).  
- The scanner provides primitives to `readRune` and `unreadRune` which wrap the `bufio.Reader` methods of the same name - maintaining positional information as they go.
- The `Scanner.Scan` method grabs the next rune and identifies it.  If it is a simple control character (i.e. parenthesis) it returns the token for that control character.  If it implies a contiguous block, it hands this work off to a specialised scan function.

The specialised scan functions make up the bulk of the work.  Mostly they are simply looping looking for a character that terminates the sequence.  `scanNumber` however is somewhat more complex.  I have chosen to handle the conflict between the use of `-` to indicate a negative number and it's use as an operator in this part of the code.  This decision makes the scanner more complicated, but simplifies things in the parser we will add later - we have to make one of these two layers somewhat "impure" so I chose to do it at the more fundamental of the two, meaning the parser can be implemented without this wart.  

Fixes: #16

### DIFF
--- a/rule/sexpr/lexer.go
+++ b/rule/sexpr/lexer.go
@@ -139,5 +139,5 @@ type ScanError struct {
 
 //
 func (se *ScanError) Error() string {
-	return se.msg
+	return fmt.Sprintf("Error:%d,%d: %s", se.Line, se.CharInLine, se.msg)
 }
--- a/rule/sexpr/lexer.go
+++ b/rule/sexpr/lexer.go
@@ -2,9 +2,11 @@ package sexpr
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"unicode"
+	"unicode/utf8"
 )
 
 type Token int
@@ -73,11 +75,12 @@ func isSymbol(r rune) bool {
 // Scanner is a lexical scanner for extracting the lexical tokens from
 // a string of characters in our rule symbolic expression language.
 type Scanner struct {
-	r             *bufio.Reader
-	byteCount     int
-	charCount     int
-	lineCount     int
-	lineCharCount int
+	r                     *bufio.Reader
+	byteCount             int
+	charCount             int
+	lineCount             int
+	lineCharCount         int
+	previousLineCharCount int
 }
 
 // NewScanner wraps a Scanner around the provided io.Reader so that we
@@ -101,12 +104,25 @@ func (s *Scanner) Scan() (Token, string, error) {
 		return LPAREN, "(", nil
 	case isRParen(rn):
 		return RPAREN, ")", nil
+	case isWhitespace(rn):
+		err := s.unreadRune(rn)
+		if err != nil {
+			return EOF, string(rn), err
+		}
+		return s.scanWhitespace()
 	}
 	return EOF, string(rn), s.newScanError("Illegal character scanned")
 }
 
 func (s *Scanner) readRune() (rune, error) {
 	rn, size, err := s.r.ReadRune()
+	// EOF is a special case, it shouldn't affect counts
+	if err == io.EOF {
+		return rn, s.eof()
+	}
+	// We need to update the counts correctly before considering
+	// any error, so that the data embedded in the ScanError is
+	// correct.
 	s.byteCount += size
 	s.charCount++
 	s.lineCharCount++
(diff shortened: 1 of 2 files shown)


rule/sexpr/lexer.go
```diff
@@ -64,3 +73,385 @@ func isComment(r rune) bool {
...
+			return NUMBER, b.String(), err
+		}
+		if rn == '-' {
+			// As we said before '-' can't appear in the
+			// body of a number, this is an error.
+			return NUMBER, b.String(), s.newScanError("invalid number format (minus can only appear at the beginning of a number)")
+		}
+
+		// Valid number parts are written to the buffer
+		if isNumber(rn) || rn == '.' {
```
asdine: If I understand correctly, this `.3..14...1....5.....9.....` won't trigger an error?

rule/sexpr/lexer.go
```diff
@@ -64,3 +73,385 @@ func isComment(r rune) bool {
...
+		if err != nil {
+			se := err.(*ScanError)
+			if se.EOF {
+				// EOF is a valid terminator for a Comment
+				break
+			}
+			return COMMENT, b.String(), err
+		}
+
+		if rn == '\n' {
```
asdine: Does this also catch `\r\n`?

rule/sexpr/lexer.go
```diff
@@ -64,3 +73,385 @@ func isComment(r rune) bool {
...
+// scanSymbol scans a contiguous block of symbol characters. Any non-symbol character will terminate it.
+func (s *Scanner) scanSymbol() (Token, string, error) {
+	var b bytes.Buffer
+
+	for {
+		rn, err := s.readRune()
+		if err != nil {
+			se := err.(*ScanError)
+			if se.EOF {
+				// EOF is a valid terminator for a Comment
```
asdine:
```suggestion
				// EOF is a valid terminator for a Symbol
```

rule/sexpr/lexer.go
tealeg: @asdine you do understand it correctly - we'll validate that it really is a valid number when we do type checking in the parser.

rule/sexpr/lexer.go
tealeg: Yes.  Because `\r` is really quite meaningless and it's always followed by a `\n`. we can just say `\r` is a white space character, but it has no impact on the line count.

asdine: But the returned string will contain `blablabla\r`, is it an issue?

rule/sexpr/lexer.go
```diff
@@ -64,3 +73,385 @@ func isComment(r rune) bool {
...
+
+		// Now we look ahead to see if a number is coming, if
+		// its anything else then this isn't a negative
+		// number, but some other form.
+		rn, err := s.readRune()
+
+		if err != nil {
+			se := err.(*ScanError)
+			if se.EOF {
+				// In reality having '-' as the final
```
YassineE: Not sure to understand this comment. 
It can be everything else, why this comment is oriented to have a `-` here?

YassineE: Oh I got it, because we first read the `-` character, then we read again and we got an EOF error. So the last character is exactly a `-`.

rule/sexpr/lexer.go
```diff
@@ -64,3 +73,385 @@ func isComment(r rune) bool {
...
+func (s *Scanner) scanBool() (Token, string, error) {
+
+	var b bytes.Buffer
+
+	for {
+		rn, err := s.readRune()
+		if err != nil {
+			se := err.(*ScanError)
+			if se.EOF {
+				// EOF is a valid terminator for a number
```
YassineE:
```suggestion
				// EOF is a valid terminator for a boolean
```

rule/sexpr/lexer_internal_test.go
```diff
@@ -144,5 +147,206 @@ func TestIsSymbol(t *testing.T) {
...
+	// actually handled by the number scan, but we'll check '-' all the same:
+	assertScanned(t, "-", "-", SYMBOL, 1, 1, 1, 1)
+}
+
+// Scanner.Scan can scan a full symbollic expression sequence.
+func TestScannerScanSequence(t *testing.T) {
+	input := `
+(and
+  (= (+ 1 -1) 0)
+  (= my-parameter "fudge sundae")) ; Crazy
```
YassineE: 😋

rule/sexpr/lexer.go
asdine: cc @tealeg

tealeg: I've made an issue to remove the embedded `\r` from the comment anyway:
https://github.com/heetch/regula/issues/41

PR #42, opened 2018-11-21
rules: Add Type, Cardinality, Term, Contract and TypedExpression

tealeg:
This PR lays further groundwork for the s-expression Parser. ( #17 )

I have defined contracts for `Expr`, `Value` and `Param` types.   These contracts will be used by the parser to perform type checking on the tree of s-expressions.  We will parse the tree recursively, when we reach a leaf-node in the tree we will check it's `Contract.ReturnType` against the expected `Type` of the `Term` it is providing for the enclosing `Expr`.  This essentially makes our language statically typed (but type is inferred), with the inherent advantage of being able to capture programming errors (e.g. calling "not" on a string) prior to evaluation.

Note that when we talk about Type here, there are abstract types understood when parsing the language, not the concrete types that appear at runtime.  Which is to say, abstractly you can require that a parameter (Term) be a `NUMBER` or  `ANY`, but in practice this will always be something concrete like a Float64 or a string. 

We may later add a "Homogeneous" flag to `Term` to allow us to specify that, for example, "eq" requires that all of its arguments be of the same type (but accepts any type).  Not however that homogeny checking will happen in the Parser not in the Contract (or at least, it would have to be triggered by the parser).

### DIFF
--- a/rule/contract.go
+++ b/rule/contract.go
@@ -1,0 +1,59 @@
+package rule
+
+// Type represents the type of a typed expression.  Any expression has
+// a return type, and some expressions also receive typed parameters.
+type Type int
+
+// These constants represent the complete set of abstract types usable
+// in expressions within Regula.  Not that these abstract types don't
+// necessary equate directly to concrete types that you might think
+// about within Go.  In particular "NUMBER" and "ANY" exist to define
+// parameters that do not have to be of a fixed type.
+const (
+	BOOLEAN Type = iota
+	STRING
+	INTEGER
+	FLOAT
+	NUMBER // A special type that can be promoted to INTEGER or FLOAT
+	ANY    // A special type that can be promoted to any other.
+)
+
+// Cardinality expresses the number of times a term might be repeated
+// in an expression.
+type Cardinality int
+
+// These constants define the complete set of cardinality possible for
+// terms of a contract.  We can expect a term to occur exactly once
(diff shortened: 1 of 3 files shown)


rule/contract.go
```diff
@@ -0,0 +1,67 @@
...
+
+// These constants define the complete set of cardinality possible for
+// terms of a contract.  We can expect a term to occur exactly once
+// (ONE), or zero, one or many times (MANY).
+const (
+	MANY = 0
+	ONE  = 1
+)
+
+type Term struct {
```
asdine: :tulip: Would be cool to have a comment for this one too, the rest is so well commented

rule/contract.go
```diff
@@ -0,0 +1,67 @@
...
+// typed, and has a defined cardinality.
+type Contract struct {
+	ReturnType Type
+	Terms      []Term
+}
+
+// A TypedExpression is an expression that declares the Type Contract
+// it makes with the context in which it appears, and with any
+// sub-expressions that it contains.  This Contract can be inspected
+// by called the Contract method of the TypedExpression interface.
```
asdine:
```suggestion
// by calling the Contract method of the TypedExpression interface.
```

rule/contract.go
tealeg: ... oops missed that one.  Resolved

PR #43, opened 2018-11-21
Add paths parameter to listing endpoint

### LINKED ISSUE #22: Ruleset listing endpoint: Add paths parameter
The `GET /ruletsets/<prefix>?list&paths` endpoint must return only the ruleset paths and not the ruleset themselves.
The `paths` option is a boolean.

The payload must look like this:

```json
{
  "revision": "2087",
  "rulesets": [
    {
      "path": "a/b/c",
    },
    {
       "path": "d/e/f",
    }
  ]
}
```

The payload is the same as without this option except that only the `path` variable is returned.

YassineE:
Fix #22

### DIFF
--- a/api/server/api.go
+++ b/api/server/api.go
@@ -75,11 +75,8 @@ func (s *rulesetAPI) list(w http.ResponseWriter, r *http.Request, prefix string)
 	}
 
 	continueToken := r.URL.Query().Get("continue")
-	if _, ok := r.URL.Query()["paths"]; ok {
-		entries, err = s.rulesets.ListPaths(r.Context(), prefix, limit, continueToken)
-	} else {
-		entries, err = s.rulesets.List(r.Context(), prefix, limit, continueToken)
-	}
+	_, ok := r.URL.Query()["paths"]
+	entries, err = s.rulesets.List(r.Context(), prefix, limit, continueToken, ok)
 
 	if err != nil {
 		if err == store.ErrNotFound {
--- a/api/server/api.go
+++ b/api/server/api.go
@@ -61,9 +61,8 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // the paths parameter is not given otherwise it fetches the rulesets paths only.
 func (s *rulesetAPI) list(w http.ResponseWriter, r *http.Request, prefix string) {
 	var (
-		entries *store.RulesetEntries
-		err     error
-		limit   int
+		err   error
+		limit int
 	)
 
 	if l := r.URL.Query().Get("limit"); l != "" {
@@ -76,7 +75,7 @@ func (s *rulesetAPI) list(w http.ResponseWriter, r *http.Request, prefix string)
 
 	continueToken := r.URL.Query().Get("continue")
 	_, ok := r.URL.Query()["paths"]
-	entries, err = s.rulesets.List(r.Context(), prefix, limit, continueToken, ok)
+	entries, err := s.rulesets.List(r.Context(), prefix, limit, continueToken, ok)
 
 	if err != nil {
 		if err == store.ErrNotFound {
--- a/api/server/api.go
+++ b/api/server/api.go
@@ -24,8 +24,7 @@ type rulesetAPI struct {
 }
 
 func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	path := strings.TrimPrefix(r.URL.Path, "/rulesets")
-	path = strings.TrimPrefix(path, "/")
+	path := strings.TrimPrefix(r.URL.Path, "/rulesets/")
 
 	if _, ok := r.URL.Query()["watch"]; ok && r.Method == "GET" {
 		ctx, cancel := context.WithTimeout(r.Context(), s.watchTimeout)
--- a/api/server/api.go
+++ b/api/server/api.go
@@ -57,7 +57,8 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusNotFound)
 }
 
-// list fetches all the rulesets from the store and writes them to the http response.
+// list fetches all the rulesets from the store and writes them to the http response if
+// the paths parameter is not given otherwise it fetches the rulesets paths only.
 func (s *rulesetAPI) list(w http.ResponseWriter, r *http.Request, prefix string) {
 	var (
 		entries *store.RulesetEntries
--- a/api/server/api.go
+++ b/api/server/api.go
@@ -60,8 +60,9 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // list fetches all the rulesets from the store and writes them to the http response.
 func (s *rulesetAPI) list(w http.ResponseWriter, r *http.Request, prefix string) {
 	var (
-		err   error
-		limit int
+		entries *store.RulesetEntries
+		err     error
+		limit   int
 	)
 
 	if l := r.URL.Query().Get("limit"); l != "" {
@@ -73,7 +74,12 @@ func (s *rulesetAPI) list(w http.ResponseWriter, r *http.Request, prefix string)
 	}
 
 	continueToken := r.URL.Query().Get("continue")
-	entries, err := s.rulesets.List(r.Context(), prefix, limit, continueToken)
+	if _, ok := r.URL.Query()["paths"]; ok {
+		entries, err = s.rulesets.ListPaths(r.Context(), prefix, limit, continueToken)
+	} else {
+		entries, err = s.rulesets.List(r.Context(), prefix, limit, continueToken)
+	}
+
 	if err != nil {
 		if err == store.ErrNotFound {
 			writeError(w, r, err, http.StatusNotFound)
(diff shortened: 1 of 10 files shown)


store/service.go
```diff
@@ -37,6 +37,8 @@ func IsValidationError(err error) bool {
 type RulesetService interface {
 	// List returns all the rulesets entries under the given prefix.
 	List(ctx context.Context, prefix string, limit int, continueToken string) (*RulesetEntries, error)
+	// ListPaths returns all rulesets path under the given prefix.
```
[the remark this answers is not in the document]
YassineE: Don't forget the suggestion feature 😉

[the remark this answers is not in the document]
YassineE: But it should be "rulesets paths", isn't it?

api/server/api.go
```diff
@@ -61,8 +60,9 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // list fetches all the rulesets from the store and writes them to the http response.
```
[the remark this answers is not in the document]
YassineE: Good catch, thank you! 👍

store/etcd/rulesets_test.go
```diff
@@ -153,6 +153,114 @@ func TestList(t *testing.T) {
 	})
 }
 
+func TestListPaths(t *testing.T) {
```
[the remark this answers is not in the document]
YassineE: I think that this is implicit? Not sure to understand clearly what do you mean?

store/etcd/rulesets.go
```diff
@@ -29,6 +30,7 @@ type RulesetService struct {
 }
 
 // List returns all the rulesets entries under the given prefix.
+// If the prefix is empty, it returns **all** the rulesets entries.
```
asdine: That's not true, it always returns a subset of the rulesets because of the limit

store/etcd/rulesets.go
```diff
@@ -89,6 +91,68 @@ func (s *RulesetService) List(ctx context.Context, prefix string, limit int, con
 	return &entries, nil
 }
 
+// ListPaths returns all rulesets paths under the given prefix.
+// If the prefix is empty, it returns **all** the rulesets paths.
+func (s *RulesetService) ListPaths(ctx context.Context, prefix string, limit int, continueToken string) (*store.RulesetEntries, error) {
```
asdine: This is almost the same function as above, do you think you can keep only one public API and specialize the work depending on the parameter you receive?

store/etcd/rulesets.go
```diff
@@ -89,6 +91,68 @@ func (s *RulesetService) List(ctx context.Context, prefix string, limit int, con
...
+	// if a prefix is provided it must always return results
+	// otherwise it doesn't exist.
+	if resp.Count == 0 && prefix != "" {
+		return nil, store.ErrNotFound
+	}
+
+	var entries store.RulesetEntries
+	entries.Revision = strconv.FormatInt(resp.Header.Revision, 10)
+	for _, pair := range resp.Kvs {
+		sk := strings.Split(string(pair.Key), "/")
```
asdine: 📖 This is not the best / simplest way to do that. Can you find a simpler way to remove the namespace from the key?

store/etcd/rulesets.go
YassineE: I've updated the comment to be more specific in https://github.com/heetch/regula/pull/43/commits/2c44b1d561f58e7a90093e8208ac35569dc337b0. Tell me if it suits you

store/etcd/rulesets_test.go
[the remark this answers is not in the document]
tealeg: Because test names have to, to an extent, be terse, it's useful to describe the desired behaviour we're testing for.  One of the roles of tests is to act as a kind of documentation for what the code *should* do.  When we have a a big test case like this, with multiple sub-tests it's probably worth spending a little extra time to summarise what it is we expect.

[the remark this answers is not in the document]
YassineE: @tealeg agreed! Thank you for the deep explanation. 🙏

store/etcd/rulesets.go
YassineE: Done in https://github.com/heetch/regula/pull/43/commits/f4c3b4cb428e0e51f9dd5ef5f0685a68cefbb540

store/etcd/rulesets_test.go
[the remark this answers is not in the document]
tealeg: OK, this is moving very much in the right direction.  One small tweak: it's not normally necessary to say in the text that "this tests...", because it is after all a test.  It's nicer to write a positive statement.  A more concrete example, say we're testing the `+` function.  We could write:

``
// TestPlus tests that plus returns the sum of two numbers.
func TestPlus(t *testing.T) {
```

.. but its better to write: 
```
// + returns the sum of two numbers
func TestPlus(t *testing.T) {
```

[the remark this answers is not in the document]
Comment (YassineE) [reply store/etcd/rulesets_test.go]: Oh okay, got it. I will try again.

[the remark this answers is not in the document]
Comment (tealeg) [reply store/etcd/rulesets_test.go]: :+1:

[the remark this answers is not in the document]
Comment (tealeg) [reply store/etcd/rulesets_test.go]: :+1:

File: store/etcd/rulesets.go (hunk 4 above)
Comment (YassineE) [reply store/etcd/rulesets.go]: 💥

File: store/etcd/rulesets.go [hunk 7]
```diff
@@ -55,7 +63,46 @@ func (s *RulesetService) List(ctx context.Context, prefix string, limit int, con
...
+	// if a prefix is provided it must always return results
+	// otherwise it doesn't exist.
+	if resp.Count == 0 && prefix != "" {
+		return nil, store.ErrNotFound
+	}
+
+	var entries store.RulesetEntries
+	entries.Revision = strconv.FormatInt(resp.Header.Revision, 10)
+	for _, pair := range resp.Kvs {
+		p := strings.TrimPrefix(string(pair.Key), path.Join(s.Namespace, "rulesets", "latest")+"/")
```
asdine:
```suggestion
		p := strings.TrimPrefix(string(pair.Key), s.latestRulesetPath("")+"/")
```

store/etcd/rulesets.go
YassineE: Good catch btw @asdine

store/etcd/rulesets.go
```diff
@@ -55,7 +63,46 @@ func (s *RulesetService) List(ctx context.Context, prefix string, limit int, con
...
 	options = append(options, clientv3.WithLimit(int64(limit)))
 
-	resp, err := s.Client.KV.Get(ctx, s.rulesetsPath(key, ""), options...)
+	if pathsOnly {
+		return s.listPaths(ctx, key, prefix, limit, options)
+	}
+	return s.listRulesets(ctx, key, prefix, limit, options)
+}
+
+func (s *RulesetService) listPaths(ctx context.Context, key, prefix string, limit int, opts []clientv3.OpOption) (*store.RulesetEntries, error) {
```
asdine: This function is the same as `listRulesets`. The only differences are:
- the options
- the complete key passed to `s.Client.KV.Get`
- what is done of a pair

How would you factorize that?

api/server/api.go
```diff
@@ -58,11 +57,13 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 }
 
-// list fetches all the rulesets from the store and writes them to the http response.
+// list fetches all the rulesets from the store and writes them to the http response if
+// the paths parameter is not given otherwise it fetches the rulesets paths only.
 func (s *rulesetAPI) list(w http.ResponseWriter, r *http.Request, prefix string) {
 	var (
-		err   error
-		limit int
+		entries *store.RulesetEntries
```
asdine: :tulip: You don't need to declare it since it will be created below on l79

store/etcd/rulesets.go
YassineE: Fixed here https://github.com/heetch/regula/pull/43/commits/0087813e243ac36457c9ddcd9acf32753a6b5787

store/etcd/rulesets.go
YassineE: ... and the generated `continue` token.

honestly, I don't see how can I do better?

asdine: You could have something like that:

```go
// line 71

if pathsOnly {
  opts = append(opts, clientv3.WithKeysOnly())
  entries, err = s.list(ctx, s.latestRulesetPath(key), opts, func(pair mvccpb.KeyValue) *store.RulesetEntry {
    p := strings.TrimPrefix(string(pair.Key), s.latestRulesetPath("")+"/")
    return &store.RulesetEntry{Path: p}, nil
  })
  if err == nil && len(entries) > 0 {
    entries.Continue = base64.URLEncoding.EncodeToString([]byte(entries[len(entries)-1].Path + "\x00"))
  }
} else {
  ...
}

if err != nil {
  return nil, blabla
}
```

asdine: Nevermind, it's not worth it because of the `continue` token. Removing it would remove a lot of code from the `listRulesets` method which makes it not interesting anymore to factorize

YassineE: Exactly. And we will end with a lot of `if pathsOnly { ... } else { ... }` blocks.

PR #44, opened 2018-11-27
Basic UI

asdine:
This PR adds a UI to Regula.
The UI is built using VueJS and Material design.
List of changes:
- New `mock` package: moved api unexported mocks to a shared package
- New `ui` package: Contains http handlers serving the UI SPA
- New `ui/app` directory: Contains the UI SPA source code
  - It displays a simple sidebar for now with the tree of rulesets
- Travis now runs the UI tests
- package `store`: Removed unnecessary `Latest` and `OneByVersion` methods

regula-ui


cmd/regula/cli/cli.go
```diff
@@ -25,8 +25,9 @@ import (
 // Config holds the server configuration.
 type Config struct {
 	Etcd struct {
-		Endpoints []string `config:"etcd-endpoints"`
-		Namespace string   `config:"etcd-namespace,required"`
+		Endpoints   []string      `config:"etcd-endpoints"`
+		Namespace   string        `config:"etcd-namespace,required"`
+		DialTimeout time.Duration `config:"etcd-dial-timemout"`
```
tealeg: Is this change strictly necessary / related?

ui/app/src/components/Sidebar.vue
```diff
@@ -0,0 +1,73 @@
...
+  const tree = {};
+
+  // turns each ruleset path in the form a/b/c into a
+  // temporary tree of this form:
+  // {a: {b: {c: []}}} and fills the tree variable
+  rulesets.forEach(({ path }) => {
+    let node = tree;
+
+    path.split('/').forEach((chunk, idx, list) => {
+      if (!Object.prototype.hasOwnProperty.call(node, chunk)) {
```
tealeg: To us civilians, this line here is worthy of a comment.

ui/app/src/components/Sidebar.vue
```diff
@@ -0,0 +1,73 @@
...
+    items: [],
+  }),
+
+  mounted() {
+    this.fetchRulesets();
+  },
+
+  methods: {
+    fetchRulesets() {
+      fetch('i/rulesets/')
```
tealeg: It almost make me just want threads ;-)

ui/handler.go
```diff
@@ -1,20 +1,109 @@
...
+
+	type response struct {
+		Rulesets []ruleset `json:"rulesets"`
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var resp response
+		var token string
+
+		// simulate a Do-While loop in Go
```
tealeg: :tulip: I think it would be more useful to indicate in your comment that you want this loop to run at least once, no matter the value of token.  Referencing a flow control construct from other languages is requiring a bit too much meta-knowledge for my tastes ;-)

cmd/regula/cli/cli.go
asdine: It is not related at all actually, it was just in my way, to be honest. About the necessity of this change, I thought it would be useful to be able to control how long someone accepts to wait for etcd to be ready I can remove it though, WDYT?

ui/app/.browserslistrc
```diff
@@ -0,0 +1,3 @@
+> 1%
+last 2 versions
+not ie <= 8
```
YassineE: The famous IE8 😂 
I have very bad memories with it.

ui/app/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "regula",
+  "version": "0.6.0",
```
YassineE: ❓ Why do we are sync between the backend version and the UI one?

YassineE: It should be `0.1.0`, isn't it?

asdine: It depends, they are part of the same project so I think that we want both projects to be synchronized, at least for now

PR #45, opened 2018-11-29
Map symbols to operations and expressions

tealeg:
This PR is in support of #17 (I've started pulling bits out of the parser branch so you can review them in small chunks). 

Add a declarative mechanism for defining the mapping between symbols in our symbolic expression language and the operator implementations in rule package.  Crucially we do this without exposing the operator implementations, thus maintaining the privacy of that package (as per the previous release).


asdine:
@tealeg ⚠️ The build is failing

rule/contract.go
```diff
@@ -70,3 +72,22 @@ type Contract struct {
 type TypedExpression interface {
 	Contract() Contract
 }
+
+// GetOperatorExpr returns an Expr that matches the provided operator
+// name. If no matching expression exists, and error will be returned.
```
asdine:
```suggestion
// name. If no matching expression exists, an error will be returned.
```

rule/contract.go
```diff
@@ -70,3 +72,22 @@ type Contract struct {
...
+		return &exprNot{}, nil
+	case "and":
+		return &exprAnd{}, nil
+	case "or":
+		return &exprOr{}, nil
+	case "in":
+		return &exprIn{}, nil
+
+	}
+	return nil, fmt.Errorf("No operator Expression called %q exists", name)
```
asdine: 🌷 🌷 Error strings should not be capitalized

tealeg: @YassineE agreed - and picking me up on not doing the right thing is a big part of what review is for.  I don't need flowers ever time you notice that I'm not good at these small things ;-)

PR #46, opened 2018-11-30
Construct operations imperitively and type / arity check as we go

tealeg:
This PR is in support of #17 

I decided to pull type and arity checking out of the parsers and implement it as part of iterative operation construction.  This simplifies the code a lot and keeps the concerns where they should be (operators worry about their own contracts, so the parser doesn't have to (aside: we should call this "Parser Care" :trollface: )). 

The intent is that the legacy Go interface for building rules should remain unchanged except for the it's new tendency to `panic` if you try to build invalid rules.   `panic` might be a bit harsh, but I am working to the assumption that nobody is going to push code out to production whilst its panicky.


rule/expr.go
```diff
@@ -10,6 +10,8 @@ import (
 // An Expr is a logical expression that can be evaluated to a value.
 type Expr interface {
 	Eval(Params) (*Value, error)
+	PushExpr(e Expr) error
```
asdine: There is an issue with the public API of this package:
- The Expr interface isn't really representative of the different behaviors we have in this package.
- `Param` and `Value` must implement `PushExpr` and `Finalise` even if they don't need them: they are not operators.
- If a user wants to create the simplest Expr possible:
  - He needs to implement 3 methods
  - Then **his code will panic because he didn't implement the `TypedExpression` interface**
  - He needs to implement 4 methods

I think we need to rework a little bit the public API by identifying:
- what is the minimum behaviour needed for a type to be integrated into this package?
- what are the different components of this package? Should Operator and Operand be part of the public API? If not, what should we provide to the user?
- how can we make sure that none of these solutions makes the code complicated to maintain?

Also, I insist on this: Don't hesitate to change anything that was there before and that is no longer relevant.

tealeg: @asdine yup, I agree.  I already have the feeling that I need to write a "How to add an operator" guide.  Packing `PushExpr` and `Finalise` onto `Expr` I think was a clear mistake, and an obvious first step is to pull `PushExpr` and `Finalise` out into an `Operator` interface.   

After that, most of the "pain" comes from the Go level usage of rules package.  We could make the parser use the Go way of doing things (i.e. use the constructor functions: `Not`, `Or`, etc..) but this will means handling panics again (We don't want those functions to have multiple returned values, otherwise the Go level expression of rules will be hideous!).

rule/contract.go
```diff
@@ -1,11 +1,29 @@
 package rule
 
-import "fmt"
-
 // Type represents the type of a typed expression.  Any expression has
 // a return type, and some expressions also receive typed parameters.
 type Type int
 
+//String returns a human readable representation of the Type.  This
```
YassineE:
```suggestion
// String returns a human readable representation of the Type.  This
```

rule/contract.go
```diff
@@ -1,11 +1,29 @@
...
+//makes Type implement the Stringer interface.
```
YassineE:
```suggestion
// makes Type implement the Stringer interface.
```

rule/error.go
```diff
@@ -0,0 +1,51 @@
...
+	MinPos   int
+	MaxPos   int
+	ErrorPos int
+}
+
+//Error returns a string representation of the ArityError.  This makes
+//ArityError implement the Error interface.
+func (ae ArityError) Error() string {
+	const template = "attempted to call %q with %d %s, but it requires %d %s"
+	argNoun := func(count int) string {
```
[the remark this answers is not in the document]
YassineE: What are you clapping here @asdine?

rule/expr.go
```diff
@@ -515,8 +545,15 @@ func (v *Value) Equal(other *Value) bool {
 	return v.compare(token.EQL, other)
 }
 
-type operander interface {
+// Operander is an interface for managing the operands of an
+// Expr that is an operation.
+type Operander interface {
+	// Operands returns of an operations Operands
```
YassineE: Not sure to understand the comment here.

rule/operator.go
```diff
@@ -0,0 +1,189 @@
...
+// the PushExpr call.
+type operator struct {
+	contract Contract
+	operands []Expr
+}
+
+// PushExpr is used to add an Expr as an operand to the operator.
+// Each call to PushExpr will check the new operand against the
+// operators Contract, in the event that the new operand does not
+// fulfil the appropriate Term of the Contract, and error will be
```
YassineE:
```suggestion
// fulfill the appropriate Term of the Contract, and error will be
```

rule/operator.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// Contract returns the operators Contract.  This makes operator
+// implement the TypedExpression interface.  Its intent is to allow
+// all operator Expr types to implement that interface (indirectly).
+func (o *operator) Contract() Contract {
+	return o.contract
+}
+
+func (o *operator) Same(c ComparableExpression) bool {
```
YassineE: Missing documentation here?

rule/operator.go
```diff
@@ -0,0 +1,189 @@
...
+				}
+			}
+			return true
+		}
+		return false
+	}
+	return false
+}
+
+//
```
YassineE: Forgot to add the comment here.

rule/error.go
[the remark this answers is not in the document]
asdine: Thinking of spelling correctly our output is important, so I clapped for the effort. Even video games sometimes don't do that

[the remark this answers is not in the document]
tealeg: Ironically there's a typo in the docstring above though ;-)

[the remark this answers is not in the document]
tealeg: .. or actually not, it's just GitHub not showing the previous line here ;-)

rule/operator.go
```diff
@@ -0,0 +1,156 @@
...
+
+func (o *operator) Eval(params Params) (*Value, error) {
+	return nil, nil
+}
+
+func (o *operator) Operands() []Expr {
+	return o.operands
+}
+
+// Support contract checking in the legacy Go interface for rule
```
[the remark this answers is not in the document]
tealeg: So, I've updated things quite radically now!

rule/contract.go
tealeg: Done.

rule/expr.go
tealeg: Nor me!  Good catch ;-)

rule/doc.go
```diff
@@ -0,0 +1,153 @@
...
+//   }
+//
+// We now need a constructor for our new type.  The main role of this
+// constructor is to initialise the new structs embedded operator
+// struct with a Contract. The Contract allows Regula to introspect
+// the Operator, and enforce the type of parameters passed to it.
+//
+// In this case our operator will have the "OpCode" "plus", it will
+// return a NUMBER (a special type that can be either float64 or
+// nit64).  The arguments we'll pass to "plus" will also be the Type
```
YassineE:
```suggestion
// int64).  The arguments we'll pass to "plus" will also be the Type
```

rule/doc.go
```diff
@@ -0,0 +1,153 @@
...
+//            case "float64":
+//                    return n.float64Sum(params)
+//            }
+//            return nil, fmt.Errorf("Invalid NUMBER type %q", typ)
+//    }
+//
+//
+// Finally, we need a way for people using language other than Go to
+// construct our Operator.  The Operator interface defines most of
+// what we need, bus we'll need to modify the GetOperator function in
```
YassineE:
```suggestion
// what we need, but we'll need to modify the GetOperator function in
```

rule/doc.go
```diff
@@ -0,0 +1,153 @@
...
+// operator.go to reflect our new operators existence.
```
YassineE: What is the relation between those two sentences: 
> Finally, we need a way for people using language other than Go to construct our Operator.

and

> The Operator interface defines most of what we need, bus we'll need to modify the GetOperator function in operator.go to reflect our new operators existence.

YassineE: Not sure to understand the meaning here.

rule/expr.go
```diff
@@ -67,29 +77,30 @@ func (n *exprNot) Eval(params Params) (*Value, error) {
...
-	}
-}
-
 type exprOr struct {
 	operator
 }
 
-// Or creates an expression that takes at least two operands and evaluates to true if one of the operands evaluates to true.
-// All the given operands must evaluate to a boolean.
-func Or(v1, v2 Expr, vN ...Expr) Expr {
```
asdine: We used to have a compile-time arity check but now we only have a runtime one, what is your reasoning about this?

tealeg: Yup, I did explicitly think about this.  It's a shame to lose the compile time check, but it complicates the implementation of these public functions and is, in fact an incomplete arity check for any type that has the `vN ...Expr` form.   For forms that don't have a variable number of operands, it's easy to maintain the compile time arity check, but for cases like this we'd have to push `v1`, `v2` and every element of `vN` onto a slice and then reexpand it inline when calling `consumeOperands`.   In an ideal world Go's compiler would let you do this:

`o.consumerOperands(v1, v2, ...vN)`

rule/doc.go
tealeg: It's reflecting that we don't have a mechanism, inherent to Go, to refer to a a Go function by a string containing its name, so we have to maintain a static mapping from an OpCode (name as a string) to an actual constructor function.  Currently it's a switch statement, in the future it'll probably need to be a map (to support manipulating it at run time).

PR #48, opened 2018-12-06
Allow the http server mux to be replaced before running the server

asdine:
This PR allows replacing the router from outside of Regula.

Previously, it was only possible to add routes to it and not replace it with another one from the outside because it was already assigned to the http server (l35).

This change assigns the router to the server only when we want to run the server (l49).

This is useful for being able to wrap the entire server and collect metrics for every request for example.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -32,7 +32,7 @@ func New(service store.RulesetService, cfg Config) *Server {
 
 	srv.logger = *cfg.Logger
 
-	srv.server = &http.Server{Handler: srv.Mux}
+	srv.server = new(http.Server)
 
 	ctx, cancel := context.WithCancel(context.Background())
 	// cancel context on shutdown to stop long running operations like watches.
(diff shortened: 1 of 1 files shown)


api/server/server.go
```diff
@@ -46,6 +46,8 @@ func New(service store.RulesetService, cfg Config) *Server {
 // Run runs the server on the chosen address. The given context must be used to
 // gracefully stop the server.
 func (s *Server) Run(ctx context.Context, addr string) error {
+	s.server.Handler = s.Mux
```
tealeg: :tulip: It's now a bit weird that we construct the Mux in `New`, and things could validly add handlers to it, only to see to be completely overidden.

asdine: Spoiler: This part will be completely restructured in the v0.6

What we want to do for the v0.5.x is the following:
- Using the default Regula server, everything works fine, all the routes are automatically setup, blablabla :rainbow: 
- If you want to customize the server though (i.e run your own flavor of the Regula server), you want to be able to do at least the following things:
  - Add your own routes
  - Wrap the entire HTTP handler with your own one, in order to add global metrics, logging etc.

You might want to replace the entire router and build your own from the ground up, but while it is possible to do so with this change, this is not officially supported.

tealeg: :+1:

PR #51, opened 2018-12-07
Baseline S-expression parser

tealeg:
This PR is another step along the road to completing #17 

Here we define the core `Parser` struct and it's main `Parse` method.  Most of the heavy lifting here is done by previous work on type and enforcement with the `rule` package, this help keep the core of the parser fairly simple (see `Parser.parseExpression`)

I've also defined `ParserError` to wrap and standardise errors of all knds that are intended to be shown the human (I am not a robot, I am a human.  Beeep.  :robot: )  submitting their code to be parsed. 

Features to follow in additional PR(s):
- Parse String, Integer and Float types.
- homogeneous type enforcement and type promotion (i.e. if the contract says MANY NUMBERS, and one of those numbers is a float, then we should convert them all to float).
- Warn the user about type promotions.


rule/sexpr/parser.go
```diff
@@ -23,3 +30,220 @@ func makeSymbolMap() *opCodeMap {
...
+// makeBoolValue extracts a BoolValue from a lexicalElement, or returns an error
+func (p *Parser) makeBoolValue(le *lexicalElement) rule.Expr {
+	if le.Literal == "true" {
+		return rule.BoolValue(true)
+	}
+	return rule.BoolValue(false)
+}
+
+// parseOperator attempts to convert the next symbol into an operator, if it fails and error is returned.
+func (p *Parser) parseOperator() (rule.Expr, error) {
```
asdine: :tulip: Shouldn't this function return a rule.Operator?
This would allow us to avoid doing multiple type assertions above (lines 166, 190, 201, 213) and only do one l217.

rule/sexpr/error.go
```diff
@@ -0,0 +1,126 @@
...
+
+const (
+	ARITY ErrorType = iota
+	TYPE
+	LEX
+	PARSE
+	OTHER
+)
+
+// String returns a human readble representation of the ErrorType.  This
```
YassineE:
```suggestion
// String returns a human readable representation of the ErrorType.  This
```

rule/sexpr/error.go
```diff
@@ -0,0 +1,126 @@
...
+// makes ErrorType implement the Stringer interface
```
YassineE:
```suggestion
// makes ErrorType implement the Stringer interface.
```

rule/sexpr/error.go
```diff
@@ -0,0 +1,126 @@
...
+		return "Lexical error"
+	case PARSE:
+		return "Parsing error"
+	case OTHER:
+		return "Error"
+	}
+	return "Undefined Error"
+}
+
+// ParserError is a wrapper for any and all errors occurring the
```
YassineE:
```suggestion
// ParserError is a wrapper for any and all errors occurring in the
```

rule/sexpr/error.go
```diff
@@ -0,0 +1,126 @@
...
+	EndByte         int
+	StartChar       int
+	EndChar         int
+	StartLine       int
+	EndLine         int
+	StartCharInLine int
+	EndCharInLine   int
+}
+
+// newParseError take a lexicalElement and an error and creates a new
```
YassineE:
```suggestion
// newParseError takes a lexicalElement and an error and creates a new
```

rule/sexpr/error.go
```diff
@@ -0,0 +1,126 @@
...
+		StartLine:       le.StartLine,
+		EndLine:         le.EndLine,
+		StartCharInLine: le.StartCharInLine,
+		EndCharInLine:   le.EndCharInLine,
+	}
+
+}
+
+// Error returns a human readable summary of the ParseError.  This
+// make ParseError implement the Error interface.
```
YassineE:
```suggestion
// makes ParseError implement the Error interface.
```

YassineE:
👍

rule/sexpr/parser.go
tealeg: I did this, and it worked a treat ;-)

rule/expr.go
```diff
@@ -320,6 +320,11 @@ type Param struct {
 	Name string `json:"name"`
 }
 
+//PushExpr implements the Expr interface, but will panic if called as Param's can't have subexpressions.
+func (v *Param) PushExpr(e Expr) {
```
asdine: Is there a way to avoid that by playing with the fact that Param and Value don't implement the Operator interface?

tealeg: Absolutely - I should have removed them when I changed the interfaces around.   Good catch.

PR #53, opened 2018-12-11
New ruleset page

asdine:
This PR adds a new page in the UI that allows creating a ruleset, this new page can be found in the `ui/app/src/views/NewRuleset/` directory.

The form is split into 3 components:

- `NewRuleset.vue`: the parent page that contains the form and manages the http call to the server
- `Signature.vue`: the component that renders the form for the path + signature of the ruleset
- `Rules.vue`: the component that renders the form for the rules

Other small changes:
- Font Awesome replaced by Material Design Icons
- Fetch replaced by Axios

### DIFF
--- a/api/client/rulesets_test.go
+++ b/api/client/rulesets_test.go
@@ -354,7 +354,7 @@ func TestRulesetService(t *testing.T) {
 		ch := cli.Rulesets.Watch(ctx, "a", "")
 		evs := <-ch
 		require.NoError(t, evs.Err)
-		require.EqualValues(t, 4, i)
+		require.EqualValues(t, 4, atomic.LoadInt32(&i))
 	})
 
 	t.Run("WatchRuleset/NotFound", func(t *testing.T) {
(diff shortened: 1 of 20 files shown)


YassineE:
Why the params are mandatory?

asdine:
@YassineE Good catch! I forgot to make them optional!

asdine:
@tealeg You're right, I'll add some comments though It's hard to comment while avoiding explaining Vue

Regarding the ACE editors, I used the default Lisp support because I didn't want to dive too much into the details in this PR. Basically, there is a way to customize ACE behaviour by adding our own so we can definitely have something coherent at some point

asdine:
:up:

YassineE:
There is no verification between the selected return type (from the dropdown list) and the result type of a rule. Is that done on purpose?

asdine:
> There is no verification between the selected return type (from the dropdown list) and the result type of a rule. Is that done on purpose?

I haven't focused on the payload sent to the server, neither on complete validation. That's for another PR

Issue #50, opened 2018-12-07
Add debug log when trying to Put a rulesets which already exists


YassineE:
It will be better if we log (debug level) when trying to add an existing rulesets.

asdine:
Can you give more details?

YassineE:
The `Put` method checks if the given rulesets has the same checksum of the latest one (if it exists).
If it exists, it won't be stored to avoid duplicates thus it worth it to add a debug log. WDYT?

YassineE:
Closing this, fixed in #55.

PR #55, opened 2018-12-12
Add Get endpoint

### LINKED ISSUE #24: New endpoint: Get a ruleset
Endpoint: `GET /rulesets/<path>`
This endpoint must return the ruleset related to the given path. By default it returns the latest version.

If the `version` (string) param is specified and non empty, the endpoint returns the specified version
Must return 404 if the path or the version doesn't exist or is malformed.

Returned payload:

```js
 {
        "path": "",
        "version": "",
        "ruleset": {}, // regula.Ruleset
        "signature": {
            "params": [{"<name>": "<type>"}],
            "returns": "<type>"
        }
        "versions": ["", ""] // list of string representing all the existing versions of the ruleset sorted by ascending order
    }
```

YassineE:
- Now all ruleset versions (UID only) are stored under a specific key like the signature for example. It eases the listing of the available versions for a ruleset.
- The `Signature` type has been exposed and moved to the `store` package.

Fix #24 & #50


YassineE:
Weird, the tests are not failing locally.

mock/store.go
```diff
@@ -25,6 +27,17 @@ type RulesetService struct {
 	EvalVersionFn    func(ctx context.Context, path, version string, params rule.Params) (*regula.EvalResult, error)
 }
 
+// Get runs GetFn if provided and increments GetCount when invoked.
```
tealeg: Somewhere (either here, or on the structure definition) we should document what `GetFn` is supposed to get - it might seem obvious from the context, but if we're allowing someone to provide this function, they should know what it is expected to do.

store/etcd/rulesets.go
```diff
@@ -36,6 +35,52 @@ func computeLimit(l int) int {
...
 
+// Get returns the ruleset related to the given path. By default, it returns the latest one.
+// It returns the related ruleset version if it's specified.
+func (s *RulesetService) Get(ctx context.Context, path, version string) (*store.RulesetEntry, error) {
+	var (
+		entry *store.RulesetEntry
+		err   error
+	)
+
+	if version == "" {
```
tealeg: This if block could be pushed out to a function with a meaningful name, that could be tested in isolation.   A small change but it would shorten the function an increase legibility.

store/etcd/rulesets.go
```diff
@@ -36,6 +35,52 @@ func computeLimit(l int) int {
...
+		entry, err = s.Latest(ctx, path)
+	} else {
+		entry, err = s.OneByVersion(ctx, path, version)
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := s.Client.KV.Get(ctx, s.versionsPath(path))
```
tealeg: :tulip: Line 55 to 66, and lines 68 to 79 differ only in the path the call to `s.Client.KV.Get` receives and a noun in the error messages.  This could be wrapped up in a function.  Whilst there might not be any other case of reuse, it again makes the flow of this function clearer, and potential stops bug duplication.

store/etcd/rulesets.go
```diff
@@ -507,18 +532,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
 	}, nil
 }
 
+// Under the key <namespace>/rulesets/entries/<p> there is the definition of the rulesets p.
```
tealeg: This should be:

```suggestion
// The key <namespace>/rulesets/entries/<p> locates the definition of the ruleset p.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +532,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
 func (s *RulesetService) rulesetsPath(p, v string) string {
 	return path.Join(s.Namespace, "rulesets", "entries", p, v)
 }
 
+// Under the key <namespace>/rulesets/checksums/<p> there is the encoded rulesets p.
```
tealeg:
```suggestion
// The key <namespace>/rulesets/checksums/<p> locates the encoded ruleset p.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +532,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
+// It is used to compare two rulesets in order to avoid duplicates.
```
tealeg:
```suggestion
// checksumsPath is used to compare two rulesets in order to avoid duplicates.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +532,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
 func (s *RulesetService) checksumsPath(p string) string {
 	return path.Join(s.Namespace, "rulesets", "checksums", p)
 }
 
+// Under the key <namespace>/rulesets/signatures/<p> there is the signature of the rulesets p.
```
tealeg:
```suggestion
// The key <namespace>/rulesets/signatures/<p> locates the signature of the ruleset p.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +532,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
 func (s *RulesetService) signaturesPath(p string) string {
 	return path.Join(s.Namespace, "rulesets", "signatures", p)
 }
 
+// Under the key <namespace>/rulesets/latest/<p> there is the path to the latest version of the rulesets p.
```
tealeg:
```suggestion
// The key <namespace>/rulesets/latest/<p> locates the path to the latest version of the ruleset p.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +532,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
 func (s *RulesetService) latestRulesetPath(p string) string {
 	return path.Join(s.Namespace, "rulesets", "latest", p)
 }
+
+// Under the key <namespace>/rulesets/versions/<p> there are all versions (UIDs part) of the rulesets p.
```
tealeg:
```suggestion
// The key <namespace>/rulesets/versions/<p> locates the list of all versions (UIDs part) of the ruleset p.
```

store/etcd/rulesets.go
YassineE: It's a good idea, but it's quite complicated to test it if it's a private method (the new one). 
Moreover, `Latest` and `OneByVersion` methods are not part of the `RulesetService` interface, thus we can't use the `mock` to assert if one or other of these methods have been called (thanks to the `*Count` fields).
If you have a suggestion, just shoot please. 🙏

YassineE:
🆙

store/etcd/rulesets.go
```diff
@@ -195,6 +241,29 @@ func (s *RulesetService) OneByVersion(ctx context.Context, path, version string)
 	return &entry, nil
 }
 
+// putVersions stores the new version or appends it to the old ones under the key key <namespace>/rulesets/versions/<path>.
```
asdine:
```suggestion
// putVersions stores the new version or appends it to the existing ones under the key <namespace>/rulesets/versions/<path>.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +563,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
 	}, nil
 }
 
+// The key <namespace>/rulesets/entries/<p> locates the definition of the ruleset p.
```
asdine:
```suggestion
// rulesetsPath returns the path where the rulesets are stored in etcd.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +563,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
 func (s *RulesetService) rulesetsPath(p, v string) string {
 	return path.Join(s.Namespace, "rulesets", "entries", p, v)
 }
 
+// The key <namespace>/rulesets/checksums/<p> locates the encoded ruleset p.
+// checksumsPath is used to compare two rulesets in order to avoid duplicates.
```
asdine:
```suggestion
```

store/etcd/rulesets.go
```diff
@@ -507,18 +563,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
 	}, nil
 }
 
+// The key <namespace>/rulesets/entries/<p> locates the definition of the ruleset p.
 func (s *RulesetService) rulesetsPath(p, v string) string {
 	return path.Join(s.Namespace, "rulesets", "entries", p, v)
 }
 
+// The key <namespace>/rulesets/checksums/<p> locates the encoded ruleset p.
```
asdine:
```suggestion
// checksumsPath returns the path where the checksums are stored in etcd.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +563,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
+// checksumsPath is used to compare two rulesets in order to avoid duplicates.
 func (s *RulesetService) checksumsPath(p string) string {
 	return path.Join(s.Namespace, "rulesets", "checksums", p)
 }
 
+// The key <namespace>/rulesets/signatures/<p> locates the signature of the ruleset p.
```
asdine:
```suggestion
signaturesPath returns the path where the signatures are stored in etcd.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +563,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
 func (s *RulesetService) signaturesPath(p string) string {
 	return path.Join(s.Namespace, "rulesets", "signatures", p)
 }
 
+// The key <namespace>/rulesets/latest/<p> locates the path to the latest version of the ruleset p.
```
asdine:
```suggestion
// latestRulesetPath returns the path where the latest version of each ruleset is stored in etcd.
```

store/etcd/rulesets.go
```diff
@@ -507,18 +563,28 @@ func (s *RulesetService) EvalVersion(ctx context.Context, path, version string,
...
 func (s *RulesetService) latestRulesetPath(p string) string {
 	return path.Join(s.Namespace, "rulesets", "latest", p)
 }
+
+// The key <namespace>/rulesets/versions/<p> locates the list of all versions (UIDs part) of the ruleset p.
```
asdine:
```suggestion
// versionsPath returns the path where the versions of each rulesets are stored in etcd.
```

store/etcd/rulesets.go
YassineE: > why are those methods public at all? They're (currently) only used inside this file and in its tests.

They are also used in the `server` package, [here](https://github.com/heetch/regula/blob/issue-24/api/server/api.go#L65) for example, this is the main usage of those methods.
This is why there is [interface](https://github.com/heetch/regula/blob/issue-24/store/service.go#L20) that exposes methods than can be used outside the package (from the `server` one for example) but if you look deeply there is some methods that are not  in it like the `Latest` for example. Once we will need it outside the package we will just have to add it in there.

tealeg: :+1:

PR #57, opened 2018-12-19
Ruleset listing endpoint: return latest version only 

### LINKED ISSUE #21: Ruleset listing endpoint: return latest version only
Currently, the endpoint `GET /rulesets/<prefix>?list` returns all the rulesets versions.
It must now only return the latest version of each ruleset.

YassineE:
To do that, I found three ways:
- retrieve the latest pointer from the [latest key](https://github.com/heetch/regula/blob/api/list-return-latest-version-only/store/etcd/rulesets.go#L515-L517) (1st call to etcd) then retrieve the ruleset with the pointer received (2nd call to etcd), thus it ends with 2 calls for each ruleset ...
- do the work above directly in a [STM](https://godoc.org/github.com/coreos/etcd/clientv3/concurrency#STM) but imo it's not the correct use case (maybe I'm wrong? if it's case please don't let me ignorant)
- change the way how the latest version is stored

I opted for the latter way, thus from now, the latest key stores directly the ruleset and not a pointer to it.

Fix #21

### DIFF
--- a/api/client/rulesets_test.go
+++ b/api/client/rulesets_test.go
@@ -354,7 +354,7 @@ func TestRulesetService(t *testing.T) {
 		ch := cli.Rulesets.Watch(ctx, "a", "")
 		evs := <-ch
 		require.NoError(t, evs.Err)
-		require.EqualValues(t, 4, i)
+		require.EqualValues(t, 4, atomic.LoadInt32(&i))
 	})
 
 	t.Run("WatchRuleset/NotFound", func(t *testing.T) {
--- a/ui/app/.eslintrc.js
+++ b/ui/app/.eslintrc.js
@@ -7,23 +7,20 @@ module.exports = {
     node: true,
     mocha: true,
   },
-  extends: [
-    'plugin:vue/essential',
-    '@vue/airbnb',
-  ],
+  extends: ['plugin:vue/essential', '@vue/airbnb'],
   rules: {
     'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
-    "import/extensions": ['error', "ignorePackages"]
+    'import/extensions': ['error', 'ignorePackages'],
+    quotes: ['error', 'single'],
   },
+
   parserOptions: {
     parser: 'babel-eslint',
   },
   settings: {
     'import/resolver': {
-      alias: [
-        ['@', path.join(__dirname, '/src')],
-      ]
-    }
-  }
+      alias: [['@', path.join(__dirname, '/src')]],
+    },
+  },
 };
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -10,8 +10,8 @@
   },
   "dependencies": {
     "@mdi/font": "^3.2.89",
+    "axios": "^0.18.0",
     "es6-promise": "^4.2.5",
-    "isomorphic-fetch": "^2.2.1",
     "roboto-fontface": "^0.10.0",
     "vue": "^2.5.17",
     "vue-router": "^3.0.1",
--- a/ui/app/package.json
+++ b/ui/app/package.json
@@ -11,10 +11,12 @@
   "dependencies": {
     "@mdi/font": "^3.2.89",
     "axios": "^0.18.0",
+    "brace": "^0.11.1",
     "es6-promise": "^4.2.5",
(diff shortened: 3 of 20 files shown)


store/etcd/rulesets.go
```diff
@@ -519,6 +512,6 @@ func (s *RulesetService) signaturesPath(p string) string {
 	return path.Join(s.Namespace, "rulesets", "signatures", p)
 }
 
-func (s *RulesetService) latestRulesetPath(p string) string {
+func (s *RulesetService) latestRuleset(p string) string {
```
asdine: You should not rename this function, this is still returning the etcd path for the latest ruleset

YassineE: Did you see that commit https://github.com/heetch/regula/pull/57/commits/85a65c96aa2516661932263c17a8d3f3fce7dc64 ?

asdine: I've just seen it but I don't understand

tealeg:
I'm not sure from the context here that I understand what we're trying to solve. What exactly was the issue with option 1?  Which of the following things are we concerned about?

- The cost of doing two reads? (Time / Load)
- The risk that the latest version changes during the read?
- The risk that the "pointer" becomes invalid during the operation (i.e. no longer resolvable)?
- Something else I haven't considered?  ;-)

If we're worried about the latest version changing whilst we read simply because we want to be sure that it really is the latest, then this is probably a "[fools errand](https://idioms.thefreedictionary.com/a+fool%27s+errand)" - there will always be border conditions, including the very large one when people load the latest version, edit it and submit it.  So the question then is, what is the real risk of getting a not up to date version? 

If we're worried about the cost of two reads, I think we might be optimising prematurely.   The approach you've put forward is essentially "denormalisation", which is a valid approach in a system where we want to optimise read performance, but I don't think we're there yet.

FWIW `STM` in etcd is really just a convenient way to bundle transactions client side - they're definitely the nicest way to ensure consistency when dealing with multiple records - I think we should use them liberally.

asdine:
Actually, the first option was 2 reads per ruleset, which means for n rulesets we got 2n calls.

The option I'm proposing is to have two calls without trying to optimize anything, first call to get a list of rulesets names and a second call to select them. Using STM or not is just a bonus.

I also agree that we shouldn't try to optimize anything and should only do so when we actually have metrics that show we must do more.

YassineE:
> WDYT?

If it's a common case to use transaction or STM for that context I am ok with this approach, especially that we don't want to waste space too much as you said. 
I thought that STM/transaction should be used exclusively when we want to do several operations (reads **and** writes) in a batch to have atomicity and avoid conflicts.

>I think we might be optimising prematurely. 

Yes maybe but if we can avoid some costs from now, it's better to act now instead of later on. WDYT?

>The option I'm proposing is to have two calls without trying to optimize anything

Why don't we do all the work in the same transaction/STM?

asdine:
> Why don't we do all the work in the same transaction/STM?

You can't do it within the same transaction or even in an STM as you can't have calls that need the result of the previous one in a transaction with etcd.

STM and etcd transactions are two different things:
- Transactions: Groups of requests that are run all at once if a certain condition is valid or not
- STM: multiple requests that make sure no writes have occurred between them. STM use transactions to do so.

Most importantly: STMs can't run a `range` request (requests that return a list of values based on a prefix)

Which means that you can use neither transaction nor STMs to run both calls at once:
- The first call must be a simple range request `.Get("prefix", WithBlabla("blabla"))`
- The second one must use a transaction to fetch all the keys returned by the first call

I don't think we have to worry about if a write occurs between both calls:
- none of the microservice we have in production worries about that
- our writes are too few
- our writes are atomic as we are using an STM to do so, which means that we'll never reach a case where we are fetching data in the middle of multiple related writes
- even if we change that and introduce multiple separated writes for the same business action, the listing is designed to work anyway:
  - the first call to the `latest/` tree will return a list ruleset names anyway
  - the second call to the `rulesets/` tree will return the list of ruleset entries anyway

TL;DR: we shouldn't worry about that at all

YassineE:
>I don't think we have to worry about if a write occurs between both calls

Indeed, to be clear I didn't want to use STM because of write operations. But just to avoid multiple round trip.

Thanks for the explanation, it's clearer now. 

About the space, we will switch to protobuf, I think we can take this in account to stay with the current solution. WDYT?

YassineE:
Closing this PR, I will reopen a new one.

Issue #23, opened 2018-11-09
Ruleset listing endpoint: Add versions parameter


asdine:
The `GET /ruletsets/<prefix>?list&versions` endpoint must return all the ruleset versions sorted by path and versions ascending
The `versions` option is a boolean.

The payload must look like this:

```json
{
  "revision": "2087",
  "rulesets": [
    {
      "path": "a/b/c",
       "ruleset": {},
      "version": "aa"
    },
    {
      "path": "a/b/c",
       "ruleset": {},
      "version": "bb"
    },
    {
       "path": "d/e/f",
       "ruleset": {},
      "version": "ac"
    }
  ]
}
```

YassineE:
The payload will look like this instead:
```
{
  "revision": "2087",
  "rulesets": [
    {
      "path": "a/b/c",
      "version": "aa"
    },
    {
      "path": "a/b/c",
      "version": "bb"
    },
    {
      "path": "d/e/f",
      "version": "ac"
    }
  ]
}
```
Without the `"ruleset"` key.

asdine:
Sorry, it was a bad copy paste from the #43 issue. I updated the issue: It is the same payload as the normal list endpoint, the difference is that now we return all the rulesets versions, not just the latest

YassineE:
What should be the behavior if the `versions` parameter is passed next to the `paths` one?

asdine:
I'd say we return a 400

YassineE:
Fixed in #60.

PR #58, opened 2018-12-24
Add ListPaths method

YassineE:
This PR adds the `ListPaths` method to the store.
The things is that the concept of listing is growing inside Regula and we now reach two axis:
- list paths only
- list rulesets (which has differents behaviour):
  - list latest rulesets version
  - list all rulesets version

To ease the legibility of the code and in order to have a source code more maintainable I suggest that we should separate both main concerns (paths vs rulesets).
It is groundwork for #23.

WDYT?


YassineE:
Closing it in favor of https://github.com/heetch/regula/pull/60.

PR #59, opened 2018-12-26
Ruleset listing endpoint: return latest version only

### LINKED ISSUE #21: Ruleset listing endpoint: return latest version only

YassineE:
This PR replaces https://github.com/heetch/regula/pull/57.
So the work is done in two steps:
- first, we retrieve the path to the latest version for each ruleset
- then, we retrieve the rulesets entries in a transaction

Fix #21.

### DIFF
--- a/store/etcd/rulesets.go
+++ b/store/etcd/rulesets.go
@@ -82,7 +82,7 @@ func (s *RulesetService) Get(ctx context.Context, path, version string) (*store.
 	return entry, nil
 }
 
-// List returns the rulesets entries under the given prefix.  If pathsOnly is set to true, only the rulesets paths will be returned.
+// List returns the latest version of each ruleset under the given prefix.  If pathsOnly is set to true, only the rulesets paths will be returned.
 // If the prefix is empty it returns entries from the beginning following the lexical ordering.
 // If the given limit is lower or equal to 0 or greater than 100, it returns 50 entries.
 func (s *RulesetService) List(ctx context.Context, prefix string, limit int, continueToken string, pathsOnly bool) (*store.RulesetEntries, error) {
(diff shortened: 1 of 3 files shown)


store/etcd/rulesets.go
```diff
@@ -164,14 +159,28 @@ func (s *RulesetService) listRulesets(ctx context.Context, key, prefix string, l
...
-	for i, pair := range resp.Kvs {
-		err = json.Unmarshal(pair.Value, &entries.Entries[i])
-		if err != nil {
-			s.Logger.Debug().Err(err).Bytes("entry", pair.Value).Msg("list: unmarshalling failed")
-			return nil, errors.Wrap(err, "failed to unmarshal entry")
+
+	for i, resps := range txnresp.Responses {
+		rr := resps.GetResponseRange()
+		for _, pair := range rr.Kvs {
+			err = json.Unmarshal(pair.Value, &entries.Entries[i])
```
asdine: I'm not sure but I think there is a mistake here, you are using the `i` of the outside loop, which means you are overwriting the same entry every time, aren't you?

store/etcd/rulesets_test.go
```diff
@@ -129,7 +129,7 @@ func TestList(t *testing.T) {
 		createRuleset(t, s, "b", rs)
 		createRuleset(t, s, "a", rs)
 
-		paths := []string{"a/1", "a", "b", "c"}
+		paths := []string{"a", "a/1", "b", "c"}
```
asdine: This shouldn't be changed lightly, this ensures the ordering is always respected. If we change that there must be a good reason to do so. What is the reasoning behind that change?

store/etcd/rulesets.go
YassineE: There is not. Because the `Responses` slice handles the responses for the whole transaction (related to the number of OpGet calls), for example if there is 3 calls to `OpGet` we will 3 entries in the slice.
Whereas the `Kvs` one handles the responses for each calls, and given that we are getting a leaf in the tree we will always have one value in the `Kvs` slice. 
I updated the code in https://github.com/heetch/regula/pull/59/commits/25e916a6b0821c48071a78ea9a8694be84e7c07a. Is it more clear?

store/etcd/rulesets_test.go
asdine: Does that mean the ordering will be different if we want to display all the versions? or if we use `pathOnly`?

YassineE: The ordering is different when the queries are based on the entries namespace.
Thus, it will be different if we list all versions. But for the `pathOnly` and the default listing (only latest version), the ordering will be the same.

YassineE:
Closing it in favor of https://github.com/heetch/regula/pull/60.

Issue #26, opened 2018-11-09
New operators: Simple arithmetic


asdine:
Addition: +
Signature: `(+ number...) -> number`
Kind: "add"

Subtraction: -
Signature: `(- number...) -> number`
Kind: "sub"

Multiplication: *
Signature: `(* number...) -> number`
Kind: "mult"

Division: /
Signature: `(/ number...) -> float`
Kind: "div"

Modulo: %
Signature: `(% int...) -> int`
Kind: "mod"

tealeg:
@asdine I not sure what the modulo of a series of more than 2 integers would be?  Every language I'm aware of only accepts to parameters to the modulo function.

Just off the top of my head I guess you could implement it as:
```
( % (% (% a b) c) d)) 
```
... but it's not clear when this would be useful.

asdine:
@tealeg Me neither actually, I suggest to go with an arity of 2, following [Wikipedia's definition](https://en.wikipedia.org/wiki/Modulo_operation):

> In computing, the modulo operation finds the remainder after division of one number by another (sometimes called modulus).

tealeg:
:+1:

PR #62, opened 2019-01-14
Simple arithmetic operations

tealeg:
Fixes: #26 

This PR adds simple arithmetic operations to the regula rules package, and provides support for them within the Regula Symbolic Expression Language.  Specifically it adds:
- `+` / `Add` (Addition for floats and integers)
- `-` / `Sub` (Subtraction for floats and integers)
- `*` / `Mult` (Multiplication for floats and integers)
- `/` / `Div' (Division for floats and integers)
- `%` / `Mod` (Modulo for integers only)

All of the above operations, except Modulo, support and arbitrarily long number of arguments and mixing of floats and integers is handled by the existing type promotion / homogenisation functionality.  Modulo is only meaningfully defined for Euclidean division of a single integer dividend by a single integer divisor, and this is reflected in our implementation. 

Along the way I made the following *significant* changes:
- Split operator expression definitions out to multiple files (they're quite verbose)
- Made mapping of rule Expression constructors to their OpCodes dynamic (allowing us to register new operations where they are defined rather than hardcoding them into the look-up function).
- Introduce a an `assert=` expression type that is only registered within a a testing environment.  This expression will test the equality of its operands and cause the enclosing golang test case to fail if they do not match.
- Write a test in `rule/sexp/parser_internal_test.go` that reads a file (`assertions.lisp`) and pushes each line in that file through a custom parser that supports the `assert=` operator.  This allows us to express our operator tests as s-expressions without reproducing Go boilerplate each time.


rule/math_expr.go
```diff
@@ -0,0 +1,393 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	for j := 1; j < len(n.operands); j++ {
+		i, err := exprToInt64(n.operands[j], params)
+		if err != nil {
+			return nil, err
+		}
+		quotient = quotient / i
```
asdine: Currently, if someone divides by 0, this line will panic.
This will translate into a 500 if the evaluation is done through HTTP, or a service crash if the evaluation is done client side and run in a separate goroutine.
This raises the question of how to deal with common errors relative to division by zero, overflow errors and other runtime errors: how do you think we should deal with those?

tealeg: Yes, I was thinking about this a little bit  - I've stopped short of doing errors and assertions about them here because the branch was already large.  But for the record, I do want to make assertions about error conditions.  The broader issue, as you say, is how we handle that at evaluation time.  We can't completely catch and eliminate those cases, and we don't yet having any branching operators that would allow the programmer to avoid them.  One idea is to have an ErrorValue and bubble this back up as the result of each evaluation, to be handled more "correctly" by the top level evaluator.  Obviously we can already return errors from "Eval" functions, but it might eventually be sensible to support some kind of error handling in the rules themselves, and having the `ErrorValue` would be a precursor for that.  

In an ideal world we wouldn't let any code that contained such errors reach production, but, again, without a way to handle it this will be hard.  We'd also need to think about how people can tests their rules without pushing them into production.

YassineE:
👏 👏 👏

PR #63, opened 2019-01-15
Move error types outside rules package

tealeg:
This PR moves some error type definition outside the rules sub-package.

These error types are used in multiple locations, and they are the source of some annoying circular import problems (and indeed some were duplicated between `regula` and `regula/rule`.  So I've pulled them out to their own subpackage to resolve this.


api/server/api.go
```diff
@@ -145,14 +145,14 @@ func (s *rulesetAPI) eval(w http.ResponseWriter, r *http.Request, path string) {
 	}
 
 	if err != nil {
-		if err == regula.ErrRulesetNotFound {
+		if err == errortype.ErrRulesetNotFound {
```
asdine: What about `errors` instead of `errortype`?

tealeg: Done

YassineE:
Nice one! 💪 👍

PR #64, opened 2019-01-15
Standalone param interface

tealeg:
To support upcoming work to introduce `let` and `if` expressions into Regula we need to add method of generating a new scoped `Params` implementation.  This work will introduce circular references if we leave the `Params` interface definition inside the `rules` package, so I've moved it out to its own package to avoid this.

### DIFF
--- a/api/client/rulesets.go
+++ b/api/client/rulesets.go
@@ -9,7 +9,7 @@ import (
 
 	"github.com/heetch/regula"
 	"github.com/heetch/regula/api"
-	"github.com/heetch/regula/param"
+	"github.com/heetch/regula/rule"
 	"github.com/pkg/errors"
 )
 
@@ -58,13 +58,13 @@ func (s *RulesetService) List(ctx context.Context, prefix string, opt *ListOptio
 
 // Eval evaluates the given ruleset with the given params.
 // It implements the regula.Evaluator interface and thus can be passed to the regula.Engine.
-func (s *RulesetService) Eval(ctx context.Context, path string, params param.Params) (*regula.EvalResult, error) {
+func (s *RulesetService) Eval(ctx context.Context, path string, params rule.Params) (*regula.EvalResult, error) {
 	return s.EvalVersion(ctx, path, "", params)
 }
 
 // EvalVersion evaluates the given ruleset version with the given params.
 // It implements the regula.Evaluator interface and thus can be passed to the regula.Engine.
-func (s *RulesetService) EvalVersion(ctx context.Context, path, version string, params param.Params) (*regula.EvalResult, error) {
+func (s *RulesetService) EvalVersion(ctx context.Context, path, version string, params rule.Params) (*regula.EvalResult, error) {
 	req, err := s.client.newRequest("GET", s.joinPath(path), nil)
 	if err != nil {
 		return nil, err
--- a/api/client/rulesets.go
+++ b/api/client/rulesets.go
@@ -9,7 +9,7 @@ import (
 
 	"github.com/heetch/regula"
 	"github.com/heetch/regula/api"
-	"github.com/heetch/regula/rule"
+	"github.com/heetch/regula/param"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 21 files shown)


tealeg:
@asdine its not really `if` but `let` (I personally feel that `if` without `let` is unworkable because of the repetition it would cause).   I'll lay out the issue, and some alternatives:

The shortest explanation of the problem:

Let requires a way to generated a new instance of something that implements `rules.Params` (so in the next PR, to add `let`, I've added a method to the interface:

```
AddParam(key string, value interface{}) (Params, error)
```
The returned `Params`  is a copy of the old one with the new symbol added.  This new one exists for the period we're evaluating  expressions inside the `let` that defined it, then it's thrown away.  Nested `lets` will result in a copy of the copy, and so on.  This is lexical scoping.

Their are multiple implementations of this interface.  Because they have to define `rules.Params` as their return type, they need to import `rules`.   We can't return the local concrete type instead because Go can't see that it implements `rules.Params`,  presumably because this would be a cyclical interface check, an infinite regression.   Up until now this is not a problem.

Here's the problem though.  Inside the `rules` package I need an implementation of `Params` to test that the `let` expression is working correctly.  If I try to use the existing, concrete implementation `regula.Params` I'll end up with an import cycle.  Placing the interface in its own package ensures that we never have this problem. 

So.. alternatives:

1. use a "mock" or testing implementation of `Params` in the tests for the `rules` package.  This is OK, but highly annoying as it'll end up being duplicate code, and that reduces the value of the tests significantly.

2. Implement `AddParrams` to modify the current `Params` rather than copy it.  Essentially we'd need to convert `Params` into a a proper stack with `push` and `pop` functionality.  This has the upside that we won't be copying maps all the time, but the downside that we'll have to deal with lookups differently.  We'd have to make this change for each implementation.  Doing lexical scoping functionally (as per my original description) is more elegant.

3. I guess moving the interface into the top level `regula` package would also work - but we'd then having a naming collision with one of the concrete implementations (incidentally the implementation that I wanted to use, but hit import cycles).

asdine:
@tealeg Thanks for the detailed answer :+1: 

2: I agree with you about lexical scoping, let's copy params rather than edit in place, we'll avoid weird bugs in upper scopes

1 & 3: I think the `regula` package happens to provide a `Param` type that is useful for other internal use cases but wasn't designed to solve them in the first place. Before considering moving it or moving the `rule.Params` type into it's own package let's explore alternative solutions, and if you still think we should move, no problem:

Sol1: For tests, import `regula.Params` in a `rule_test` package instead of the `rule` package. The `rule_test` package will be considered by Go as a different package and thus we shouldn't have import cycles.

Sol2: Generate an "official" mock for `rule.Params` and place it in the `mock` package. This has the advantage of decoupling `regula.Params` from the rest of the codebase but also we could add more features to the mock if needed. We will of course copy paste the `regula.Params` code first but a wise man once said `A little bit of copying is better than a little dependency` (in a totally different context yes, but still)

tealeg:
> @tealeg Thanks for the detailed answer +1

T'was a pleasure!
 
> 2: I agree with you about lexical scoping, let's copy params rather than edit in place, we'll avoid weird bugs in upper scopes

:+1: great.  Done deal.

> 
> 1 & 3: I think the `regula` package happens to provide a `Param` type that is useful for other internal use cases but wasn't designed to solve them in the first place. Before considering moving it or moving the `rule.Params` type into it's own package let's explore alternative solutions, and if you still think we should move, no problem:

Yes, I agree.  The problem is that the `regula.Params` implementation is exactly the one you'd want to use in testing.  The only reason for multiple implementations I can see is that the API version using `string` as it's native format rather than `interface{}`

> 
> Sol1: For tests, import `regula.Params` in a `rule_test` package instead of the `rule` package. The `rule_test` package will be considered by Go as a different package and thus we shouldn't have import cycles.

This is a great solution.  So much so that I even did that initially.  The problem comes when you want to do internal testing.  Purely as an example -  in the `let` branch I introduced code to the `operator` to allow marking one or more `Terms` as being body forms.  When you have such `Terms` you set the return type of the operator to match that of the body form.   The method that does this work is private on the operator.  I could invoke it by calling `Finalise` but that would trigger homogenisation, which could obscure errors in the code I want to test. 

> 
> Sol2: Generate an "official" mock for `rule.Params` and place it in the `mock` package. This has the advantage of decoupling `regula.Params` from the rest of the codebase but also we could add more features to the mock if needed. We will of course copy paste the `regula.Params` code first but a wise man once said `A little bit of copying is better than a little dependency` (in a totally different context yes, but still)

Yes.  I think this might have to be the way in the end.  I'll take this route, and park this PR.   Give me a :+1: if you agree ;-)

tealeg:
OK. decided to go with mocking after all.

PR #65, opened 2019-01-15
Add a Let form

tealeg:
Whilst adding the arithmetic operations it became clear that we will need a branching construct (e.g. `if`) to avoid some classes of error at runtime (so rules can guard against divide-by-zero, for example).

Once you introduce `if` (which I'll do in the next PR), you quickly hit the case where you have to repeat sections of your code, unless you have a way to store a variable value in line.  So, the pressure to introduce the `let` form is squarely there - without it rule authors will have to duplicate code (possibly introducing errors) and we'll have to evaluate it twice. 

To avoid the complexity of introducing a second namespace for variables, this PR instead allows for scoped parameter sets.  As we recursively walk the tree, both at parse time and at evaluation time, we might encounter a `let` statement.  When this occurs we create a new copy of our `Params` struct (both kinds, parse time and eval time) and shove a new parameter into it.  When we're done parsing/evaluating with that scope we'l throw away the new `Params` and go back to the previous one.

I've also arbitrarily added the rule that you can't overwrite existing parameter values in the new scope (i.e you can't use a name that already used by a parameter). 

One other thing.  To avoid having to deal with lists that don't start with an operator, we strictly limit our `let` statement to binding one new symbol per `let`.  So in practice our `let` looks like this:

```
(let SYMBOL VALUE-FORM BODY)
```

... for example:

```
(let my-variable (+ 1 1) (* 2 my-variable))
```
FWIW, this form will return `Int64Value( 4)`

There's quite a lot involved in making this change, but it really is the most succinct way of doing it.  Key things to understand:
- `parseExpression` now knows not to evaluate the symbol in position 1 of a `let` statement, but rather to parse the next expression (the value form) and create a new parameter scope with a new Param type of the same type as the value form bound to the symbol.  Once its done this both the new Param and the value form are pushed to the operator as Exprs in the normal way (we need them again at Eval time!)
- When we finalise operators we now take account of a concept in the Contract called "IsBody" - an operator might have multiple body forms.  All of these forms must have the same return type, and this return type will become the return type of the operator if the operator defines its return type as `ANY`.  This is compatible with homogenisation where that applies.

Again.. this might be complex, so I'm happy to walk anyone through this if required.

### DIFF
--- a/rule/expr.go
+++ b/rule/expr.go
@@ -5,11 +5,13 @@ import (
 	"fmt"
 	"go/token"
 	"strconv"
+
+	"github.com/heetch/regula/param"
 )
 
 // An Expr is a logical expression that can be evaluated to a value.
 type Expr interface {
-	Eval(Params) (*Value, error)
+	Eval(param.Params) (*Value, error)
 	Contract() Contract
 }
 
@@ -18,18 +20,6 @@ type Expr interface {
 type ComparableExpression interface {
 	Same(ComparableExpression) bool
 	GetKind() string
-}
-
-// A Params is a set of parameters passed on rule evaluation.
-// It provides type safe methods to query params.
-type Params interface {
-	GetString(key string) (string, error)
-	GetBool(key string) (bool, error)
-	GetInt64(key string) (int64, error)
-	GetFloat64(key string) (float64, error)
-	Keys() []string
-	EncodeValue(key string) (string, error)
-	AddParam(key string, value interface{}) (Params, error)
 }
 
 // Param is an expression used to select a parameter passed during evaluation and return its corresponding value.
@@ -118,7 +108,7 @@ func Float64Param(name string) *Param {
 }
 
 // Eval extracts a value from the given parameters.
-func (p *Param) Eval(params Params) (*Value, error) {
+func (p *Param) Eval(params param.Params) (*Value, error) {
 	if params == nil {
 		return nil, errors.New("params is nil")
 	}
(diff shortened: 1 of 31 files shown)


rule/expr.go
```diff
@@ -29,13 +36,7 @@ type Params interface {
...
 	Keys() []string
 	EncodeValue(key string) (string, error)
-}
-
-// Param is an expression used to select a parameter passed during evaluation and return its corresponding value.
-type Param struct {
-	Kind string `json:"kind"`
-	Type string `json:"type"`
-	Name string `json:"name"`
+	AddParam(key string, value interface{}) (Params, error)
```
asdine: I was thinking maybe we should consider parameters and variables as a stack (only in the `rule` package).
`context.Context` elegantly uses embedding to implement a stack without copy, maybe we could do something like that?

Currently, all the `Params` methods are for reading values; for writes we could do something like that:

```go
type paramStack struct {
  Params Params

  key string
  value interface{}
}

func (p paramStack) GetFloat64(key string) (float64, error) {
  if p.key == key {
    return p.value.(float64), nil
  }

  return p.Params.GetFloat(key)
}

...

func addParam(params Params, key string, value interface{}) Params {
  return &paramStack{
    Params: params,
    Key: key,
    Value: value,
  }
}
```

That way, we have the following benefits:
- we don't have to implement an AddParam method to the `Params` interface
- we have the certitude params are not mutated during evaluation, since we are not depending on the their underlying `AddParam` implementation but ours
- we avoid copy
- upper scopes don't even know new variables have been created in lower scopes

WDYT?

tealeg:
@asdine :up:

Issue #27, opened 2018-11-09
New operators: Comparison


asdine:
Comparison operators must compare **the same types only**

Greater than: >
Signature: `(> any...) -> boolean`
Kind: "gt"

Greater than or equal: >=
Signature: `(>= any...) -> boolean`
Kind: "gte"

Lesser than: <
Signature: `(< any...) -> boolean`
Kind: "lt"

Lesser than or equal: <=
Signature: `(<= any...) -> boolean`
Kind: "lte"

PR #67, opened 2019-01-22
Implement Less-than operator (<)

tealeg:
Partial fix for #27 

As per the spec, this implementation handles all fundatmental types - even though that's a little weird (what does it really mean to say that `#false` is  less that `#true`?)

Another point of discussion: equality is defined, per type, in `expr.go`, but I have put the per-type definitions here next to the operator definitions.  It would be a small matter to move them, but I thought I'd wait and have the chat about type coverage first.

### DIFF
--- a/rule/eq_expr.go
+++ b/rule/eq_expr.go
@@ -147,6 +147,28 @@ func (n *exprLT) floatLT(params Params) (*Value, error) {
 	return BoolValue(true), nil
 }
 
+// perform a less-than comparison on a sequence of strings
+func (n *exprLT) stringLT(params Params) (*Value, error) {
+	var s0, s1 string
+	var err error
+
+	s0, err = exprToString(n.operands[0], params)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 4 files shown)


rule/eq_expr.go
```diff
@@ -62,3 +68,152 @@ func (n *exprEq) Eval(params Params) (*Value, error) {
...
+	i0, err = exprToInt64(n.operands[0], params)
+	if err != nil {
+		return nil, err
+	}
+	for j := 1; j < len(n.operands); j++ {
+		i1, err = exprToInt64(n.operands[j], params)
+		if err != nil {
+			return nil, err
+		}
+		if i0 >= i1 {
```
YassineE: Given that you are implementing the LT symbol, the check should be `>` instead of `>=`, isn't it?
And it will be `>=` for the LTE one.

tealeg: Not that the next line returns `BoolValue(false)` - It's `>=` here because the only true case is `<`.

YassineE: Oh right, my bad!

PR #73, opened 2019-01-25
Add <= and >= operators 

tealeg:
The final part of the fix for #27 

Add operators for `<=` and `>=` comparisons.

Initially I implemented these in terms of the existing `=`, `<` and `>` operators, but this was pretty clunky and would have been slow, so I went back and wrote them from scratch.


asdine:
> Maybe transfer this thought into an issue?

Yes that's what I should have done from the beginning :+1:

PR #74, opened 2019-01-30
Fix sidebar

asdine:
This PR improves how the sidebar works, especially to allow assigning the right url to the right node.
Leaf nodes are now clickable and redirect to the right page (even though that page is empty)
Non-leaf nodes automatically expand to display the subtree


ui/app/src/views/LatestRuleset/LatestRuleset.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+  <div>
+    YOOOO
```
tealeg:
```suggestion

```

PR #81, opened 2019-02-12
Percentile operator

### LINKED ISSUE #80: New operators: Percentile
The percentile operator can establish whether the hash of a given value is within a given percentile.  It is intended to be used to alot items to an experimental grouping.  Percentile takes two parameters (ANY, INTEGER) and returns a Boolean:

The hash of the ANY value modulo 100 gives us its result.   If the result is <= to the provided INTEGER value, we consider it to be within that percentile.

tealeg:
Fixes #80 
Add a `percentile` operator in-line with the specification.

### DIFF
--- a/rule/group_expr.go
+++ b/rule/group_expr.go
@@ -1,5 +1,9 @@
 package rule
 
+func init() {
+	Operators["percentile"] = func() Operator { return newExprPercentile() }
+}
+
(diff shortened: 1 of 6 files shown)


YassineE:
Don't forget to rebase this since we merge #79.

PR #83, opened 2019-02-13
Float to int

### LINKED ISSUE #82: New operators: float->int
We need an operator to explicitly cast a float to an integer.

tealeg:
Fixes #82 

This PR introduces the `float->int` operator, and along the way adds a missing test for `int->float`.

### DIFF
--- a/rule/cast_expr.go
+++ b/rule/cast_expr.go
@@ -6,7 +6,12 @@ import (
 
 func init() {
 	Operators["intToFloat"] = func() Operator { return newExprIntToFloat() }
+	Operators["floatToInt"] = func() Operator { return newExprFloatToInt() }
 }
+
+/////////////////////////
+// IntToFloat Operator //
+/////////////////////////
 
 // exprIntToFloat is an operation that converts an Int64 type to a Float64 type.
 type exprIntToFloat struct {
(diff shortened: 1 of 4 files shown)


rule/cast_expr.go
```diff
@@ -51,3 +56,52 @@ func (n *exprIntToFloat) Eval(params Params) (*Value, error) {
...
+						Cardinality: ONE,
+						Type:        FLOAT,
+					},
+				},
+			},
+		},
+	}
+}
+
+// FloatToInt is an operator that converts as Float64 into an Int64 type.
```
YassineE:
```suggestion
// FloatToInt is an operator that converts a Float64 into an Int64 type.
```

rule/cast_expr.go
```diff
@@ -51,3 +56,52 @@ func (n *exprIntToFloat) Eval(params Params) (*Value, error) {
...
+func FloatToInt(vN ...Expr) Expr {
```
YassineE: I am wondering why are you passing a list of Expr whereas the cardinality is one?

tealeg: Ah.. quite right, I'll fix that.

YassineE: The same goes for the `IntToFloat` function.

tealeg: Done.

tealeg:
@YassineE :up:

Issue #84, opened 2019-02-15
Pretty-print rules as RUSE syntax


tealeg:
We need the ability to convert expression back into RUSE format so that we can display current rules in the UI.  This is a prerequisite to doing the UI to display rule versions.

PR #85, opened 2019-02-15
Pretty print rules as RUSE syntax

tealeg:
This PR fixes issue #84 and is a prerequisite of all UI work displaying existing rules.
Implent the `PrettyPrint` function for regula expressions, essentially dumping them back as a rigidly formatted RUSE syntax.

Note that we handle `let` and `if` forms differently from other operators.  

Pretty printing is a complex and annoying thing, so please point out anywhere this code needs to be better documented.

### DIFF
--- a/rule/sexpr/print.go
+++ b/rule/sexpr/print.go
@@ -1,17 +1,129 @@
 package sexpr
 
 import (
+	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/heetch/regula/rule"
 )
 
-func Print(e rule.Expr) (string, error) {
+var sm *opCodeMap
+
+func init() {
+	sm = makeSymbolMap()
+}
+
+// prettyPrintOp is the generic printer for operations
+func prettyPrintOp(indent, wrap int, e rule.Expr) (string, error) {
+	output := ""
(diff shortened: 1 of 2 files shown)


YassineE:
What is the "RUSE" syntax?

tealeg:
RUSE = RUle Symbolic Expressions. It's the "Lispy" DSL's name.

rule/sexpr/print.go
```diff
@@ -0,0 +1,164 @@
...
+			// go back via a real float64 for printing.
+			f, err := strconv.ParseFloat(v.Data, 64)
+			if err != nil {
+				return "", err
+			}
+			line = strconv.FormatFloat(f, 'f', -1, 64)
+		case "bool":
+			line = "#" + v.Data
+		case "string":
+			line = fmt.Sprintf("%q", v.Data)
```
asdine:
```suggestion
			line = string(v.Data)
```

tealeg: @asdine that's not a valid change.  The string needs to be quoted in the output.

asdine: Oh ok, makes sense :+1:

PR #87, opened 2019-02-20
Introduce Signature as a top level type

asdine:
This PR is the first of several PRs that will decouple the ruleset from its signature.
In this one, the most important change is that the `Ruleset` no longer has a `ReturnType`, it now has a `Signature` that describes the return type, the param names, and param types.

The `Signature` has been written as a DSL because it will mostly be used within tests and/or to build a `Ruleset`, so I wanted to provide something easy to build.

Also, `Engine` methods no longer return an `EvalResult`. I discussed with @sixstone-qq and he told me that since this variable was never used, that most of the time the version would not be important.
So I decided to remove that return value and instead add a simple untyped `Get` method that returns an `EvalResult` and let users decode it the way they want. I added some helpers to do that (`EvalResult.ToInt64()`, `EvalResult.ToString()`, etc.)

This PR won't be merged until upcoming pull requests are validated and merged against it (so it's normal if the build fails until then)


YassineE:
What is the plan for the next coming PRs?

asdine:
> What is the plan for the next coming PRs?

@YassineE Changing the way we rulesets are created in the store. The signature will be created first, then rules, not both at the same time.

asdine:
@tealeg :up:

engine.go
```diff
@@ -156,6 +158,26 @@ type EvalResult struct {
 	Version string
 }
 
+// ToString decodes the value as a string.
```
tealeg: Seems like there's a strong overlap between this code (all of these `To...()` methods) and the private `exprTo....()` functions in `rule/expr.go`.  I wonder if we should just support a `To<go type>()` method on `rule.Value` types instead of doing these conversions in multiple places?

ruleset.go
```diff
@@ -127,15 +98,76 @@ type Signature struct {
...
+	return &Signature{
+		ParamTypes: make(map[string]string),
 	}
+}
 
-	return &Signature{
-		ParamTypes: pt,
-		ReturnType: rs.Type,
+// StringP adds a string param to the signature.
+func (s *Signature) StringP(name string) *Signature {
```
tealeg: My inner lisper automatically assumed "StringP" would return a boolean :-)   P is a common short-hand for "predicate", and in Lisp that generally means type predicate and thus `(string-p "hello world") -> true`

ruleset.go
```diff
@@ -127,15 +98,76 @@ type Signature struct {
 	ParamTypes map[string]string
 }
 
-// NewSignature returns the Signature of the given ruleset.
-func NewSignature(rs *Ruleset) *Signature {
-	pt := make(map[string]string)
-	for _, p := range rs.Params() {
-		pt[p.Name] = p.Type
+// NewSignature create
+func NewSignature() *Signature {
```
tealeg: Why don't we just pass in the return type as a constant instead of having all the methods to set it?   I'm not sure what the advantage is?  We always need a return type for it to be a valid signature.

ruleset.go
tealeg: For clarity - I'm not asking you to do anything here, just recording my surprise ;-)

PR #88, opened 2019-02-21
Split etcd files

asdine:
This PR simply splits etcd `rulesets.go` file into multiple files


YassineE:
@asdine Can we merge this?

PR #90, opened 2019-02-27
Decouple Signature from Ruleset

asdine:
This PR significantly changes how Regula deals with Signatures and Rulesets:

## Library: Rulesets & Signatures

Here we are talking about basic types that Regula as a library provides:

- A Ruleset is now a simple set of rules. It has no additional constraints regarding return or param types.
- A Signature defines a return type and params.

These are primitives that are made to be used independently or combined at will.

## API & Storage

Here we are talking about Regula as a service that makes sure rulesets follow signatures.
In the eye of an API user, a ruleset is a bunch of rules that follow a certain signature.

:warning: This PR only deals with the storage part (that's why the build still fails)

The RulesetService now requires that the user creates a signature first (see new `Create` method) before being able to `Put` a ruleset.

When the user `Put`s a ruleset, it gets validated against the stored signature.

This allows creating a signature with all the parameters you want even if you don't want to use them right away.

Next PR will simply fix the breaking changes in the API and the HTTP client, and add a new endpoint for creating signatures (probably `POST /rulesets/<path>`)


store/etcd/get.go
```diff
@@ -13,113 +13,63 @@ import (
...
+	if version == "" {
+		ops = append(ops, clientv3.OpGet(s.rulesetsPath(path, "")+versionSeparator, clientv3.WithLastKey()...))
+	} else {
+		ops = append(ops, clientv3.OpGet(s.rulesetsPath(path, version)))
 	}
-	entry.Versions = v.Versions
-
-	resp, err = s.Client.KV.Get(ctx, s.signaturesPath(path))
+	// running all the requests within a single transaction so only one network round trip is performed.
+	resp, err := s.Client.KV.Txn(ctx).Then(ops...).Commit()
```
YassineE: 👏

store/etcd/get.go
```diff
@@ -13,113 +13,63 @@ import (
...
-// It returns store.ErrNotFound if the path doesn't exist or if it's not a ruleset.
-func (s *RulesetService) OneByVersion(ctx context.Context, path, version string) (*store.RulesetEntry, error) {
-	if path == "" {
-		return nil, store.ErrNotFound
-	}
-
-	resp, err := s.Client.KV.Get(ctx, s.entriesPath(path, version))
+	// decode ruleset
+	var ruleset pb.Ruleset
+	err = proto.Unmarshal(resp.Responses[2].GetResponseRange().Kvs[0].Value, &ruleset)
```
YassineE: If we just have a signature without rulesets, it will panic here, the same above for the versions.
I understood that you make the verification once for the whole entities (L.43) but it is not relevant if we just have a signature without a rulesets "added" to it.

store/etcd/proto.go
```diff
@@ -168,16 +163,24 @@ func valueFromProtobuf(v *pb.Value) *rule.Value {
 	}
 }
 
+// signatureToProtobuf creates a protobuf Signature from a regula.Signature.
+func signatureToProtobuf(sig *regula.Signature) *pb.Signature {
+	return &pb.Signature{
+		ReturnType: sig.ReturnType,
+		ParamTypes: sig.ParamTypes,
```
YassineE: If the signature is updated (especially the `ParamTypes` map) after being passed to this function, the `pb.Signature` will also be updated. Do we really want this?

store/etcd/proto.go
```diff
@@ -168,16 +163,24 @@ func valueFromProtobuf(v *pb.Value) *rule.Value {
...
 	}
 
-	for k, v := range s.ParamTypes {
-		sig.ParamTypes[k] = v
+	if sig.ParamTypes == nil {
+		sig.ParamTypes = make(map[string]string)
 	}
 
-	return sig
+	return &sig
```
YassineE: 📖 What is the difference between:
```golang
sig := &regula.Signature{...}
...
return sig
```

And
```golang
sig := regula.Signature{...}
...
return &sig
```

store/etcd/put.go
```diff
@@ -3,236 +3,156 @@ package etcd
...
-		return nil, errors.Wrap(err, "failed to put ruleset")
+	// update checksum if rules have changed
+	changed, err := p.updateChecksum(p.stm, path, data)
+	if err != nil {
+		return nil, err
 	}
 
-	return &entry, err
-}
+	if !changed {
```
YassineE: Maybe we should do this firstly, that case, if the ruleset hasn't changed, we return it directly without checking if the ruleset satisfies the signature.

store/etcd/put.go
```diff
@@ -3,236 +3,156 @@ package etcd
...
+	p.stm.Put(p.s.latestVersionPath(path), p.s.rulesetsPath(path, entry.Version))
+
+	return &entry, p.updateVersionRegistry(p.stm, path, entry.Version)
 }
 
-func validateRuleset(path string, rs *regula.Ruleset) (*regula.Signature, error) {
-	err := validateRulesetName(path)
-	if err != nil {
-		return nil, err
+// validateRules fetches the signature from the store and validates all the rules against it.
```
YassineE:
```suggestion
// validateRuleset fetches the signature from the store and validates all the rules against it.
```

store/etcd/put.go
```diff
@@ -3,236 +3,156 @@ package etcd
...
+	data := stm.Get(p.s.signaturesPath(path))
+	if data == "" {
+		return nil, store.ErrSignatureNotFound
 	}
 
-	sig := regula.NewSignature(rs)
+	var pbsig pb.Signature
+	err := proto.Unmarshal([]byte(data), &pbsig)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to decode signature")
```
YassineE: Usually we use `unmarshall` instead of `decode`. 
I am not saying that this one is better than the other but we should agree on which one we use.

store/etcd/put.go
```diff
@@ -3,236 +3,156 @@ package etcd
...
-			Field:  "path",
-			Value:  path,
-			Reason: "invalid format",
-		}
+// updateChecksum generates a checksum from the given data and stores it if it has changed.
+// It returns a boolean that is true if the checksum has changed.
+func (p *rulesPutter) updateChecksum(stm concurrency.STM, path string, data []byte) (bool, error) {
+	// generate a checksum from the rules for comparison purpose
+	h := md5.New()
+	_, err := h.Write(data)
```
YassineE: 👏 No more `json` usage! 😎

store/etcd/rulesets.go
```diff
@@ -18,31 +19,47 @@ type RulesetService struct {
 	Namespace string
 }
 
-// entriesPath returns the path where the rulesets are stored in etcd.
-func (s *RulesetService) entriesPath(p, v string) string {
+// rulesPath returns the path where the rulesets are stored in etcd.
```
YassineE:
```suggestion
// rulesetsPath returns the path where the rulesets are stored in etcd.
```

store/service.go
```diff
@@ -3,32 +3,35 @@ package store
...
+	ErrSignatureNotFound    = errors.New("signature not found")
 	ErrInvalidContinueToken = errors.New("invalid continue token")
+	ErrAlreadyExists        = errors.New("already exists")
 )
 
 // RulesetService manages rulesets.
 type RulesetService interface {
-	// Get returns the ruleset related to the given path. By default, it returns the latest one.
-	// It returns the related ruleset version if it's specified.
+	// Create a ruleset entry using a signature.
```
YassineE: I am not fan of this comment. Maybe we should specify that it creates a "bare" ruleset?

store/etcd/get.go
tealeg: https://giphy.com/gifs/vh1-l1J9I0WUCkL0YfumI

store/etcd/put.go
```diff
@@ -3,236 +3,156 @@ package etcd
...
-		}
-		version := k.String()
+	return entry, err
+}
 
-		err = s.putVersions(stm, path, version)
-		if err != nil {
-			return err
-		}
+// rulesPutter is responsible for validating and storing rules, updating checksums and other actions
```
tealeg: :tulip: I like the `rulesPutter`, but I prefer to use the `rulesWood` or the `rules9Iron`.  :golf:

store/etcd/proto.go
asdine: It's the same thing

store/etcd/put.go
asdine: Unmarshal is just more specific than decode, what we are actually doing is decoding, it is a more common word than unmarshal.

store/etcd/proto.go
asdine: It makes sense, it's just that for now, it's useless to copy it just in case.

PR #91, opened 2019-02-27
Add Create endpoint

asdine:
This PR adds a new endpoint:

```
POST /rulesets/<path>
```

This endpoint receives a signature and creates a new empty ruleset entry in the store.


api/client/rulesets.go
```diff
@@ -111,6 +111,19 @@ func (s *RulesetService) Put(ctx context.Context, path string, rs *regula.Rulese
 	return &resp, err
 }
 
+// Create creates a ruleset using on the given path using the given signature.
```
tealeg: This docstring needs attention, it doesn't make any sense at the moment.

tealeg:
```suggestion
// Create creates a ruleset at the given path using the given signature.
```

asdine:
:up: @tealeg

PR #93, opened 2019-03-04
Use http.FileSystem interface

YassineE:
Use [http.FileSystem](https://golang.org/pkg/net/http/#FileSystem) when creating ui handler. It allows using a file system from outside Regula.

### DIFF
--- a/ui/handler.go
+++ b/ui/handler.go
@@ -41,6 +41,7 @@ func NewHandler(service store.RulesetService, fs http.FileSystem) http.Handler {
 			writeError(w, r, err, http.StatusInternalServerError)
 			return
 		}
+		defer f.Close()
 		_, err = io.Copy(w, f)
 		if err != nil {
 			writeError(w, r, err, http.StatusInternalServerError)
--- a/ui/handler.go
+++ b/ui/handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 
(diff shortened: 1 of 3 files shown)


ui/handler.go
```diff
@@ -22,22 +21,19 @@ var (
...
-
+	h := http.FileServer(fs)
+	mux.Handle("/css/", h)
+	mux.Handle("/js/", h)
+	mux.Handle("/fonts/", h)
 	// catch all url that deleguates the routing to the front app router
-	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		http.ServeFile(w, r, filepath.Join(distPath, "index.html"))
-	})
+	mux.Handle("/", h)
```
asdine: The catchall URL must serve only the `index.html` file and nothing else, otherwise you are allowing people to access other files that are contained in the same directory as your dist path

YassineE: The `index.html` with the others `*.html` files (if there is) and the `css/`, `js/`,... folders. Isn't it?

YassineE:
@asdine 🆙 
@tealeg re-requested review from you.

ui/handler.go
```diff
@@ -22,21 +22,30 @@ var (
...
+	mux.Handle("/fonts/", h)
 
 	// catch all url that deleguates the routing to the front app router
 	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
-		http.ServeFile(w, r, filepath.Join(distPath, "index.html"))
+		f, err := fs.Open("index.html")
+		if err != nil {
+			writeError(w, r, err, http.StatusInternalServerError)
+			return
+		}
```
asdine: The file must be closed once you're done with it

```suggestion
		}
		defer f.Close()
```

PR #94, opened 2019-03-13
Decouple Signature from Ruleset

asdine:
This PR changes the way Regula deals with rulesets.

In order to create a ruleset, a client must now define a *Signature*. This signature is immutable.

Here are the list of big changes that this PR brings:

## HTTP API

- New endpoint `POST /rulesets`: Create a ruleset.
Payload: Signature
- `api.Ruleset` has been removed in favor of `regula.Ruleset`

## RulesetService

- New `Create` method used to create a ruleset using a signature. This method is called by the new HTTP endpoint described above
- `Put` now only takes a slice of rules:
Before: `Put(ctx context.Context, path string, rs *regula.Ruleset) (*RulesetEntry, error)`
After: `Put(ctx context.Context, path string, rules []*rule.Rule) (*regula.Ruleset, error)`
- `store.RulesetEntry` has been removed, methods that used to return that type now return a `regula.Ruleset`

## ETCD

- `ruleset.go` has been split into multiple files
- Data is now organized like this:

| Data | Path in tree |
| --- | --- |
| signatures | `<namespace>/rulesets/signatures/<path>` |
| rules | `<namespace>/rulesets/rules/<path>` |
| checksums | `<namespace>/rulesets/cheksums/<path>` |
| latest version string of a given ruleset | `<namespace>/rulesets/latest/<path>` |
| list of all the versions for a given ruleset | `<namespace>/rulesets/versions/<path>` |

## Engine

- The engine no longer returns an `EvalResult` for the `Get*` methods
- A new `Get` method returns a raw `EvalResult`
- The `Evaluator` now contains only one method `Eval`, `EvalVersion` has been removed

## Rule package

- `Value` contains a list of methods for conversion to selected types (`ToString`, `ToInt64`, etc.). These methods are used everywhere we used to manually do the conversion.

:warning: The build is failing because the `ui/` package needs a special treatment, especially on the front side. It will be done in another PR


store/etcd/rulesets.go
```diff
@@ -18,31 +19,47 @@ type RulesetService struct {
 	Namespace string
 }
 
-// entriesPath returns the path where the rulesets are stored in etcd.
-func (s *RulesetService) entriesPath(p, v string) string {
+// rulesPath returns the path where the rulesets are stored in etcd.
```
YassineE:
```suggestion
// rulesPath returns the path where the rules of rulesets are stored in etcd.
```

store/service.go
```diff
@@ -3,36 +3,37 @@ package store
...
-	// Get returns the ruleset related to the given path. By default, it returns the latest one.
-	// It returns the related ruleset version if it's specified.
-	Get(ctx context.Context, path, version string) (*RulesetEntry, error)
-	// List returns the latest version of each ruleset under the given prefix.
-	// If the prefix is empty, it returns entries from the beginning following the lexical order.
-	// The listing can be customised using the ListOptions type.
-	List(ctx context.Context, prefix string, opt *ListOptions) (*RulesetEntries, error)
+	// Create a ruleset entry using a signature.
+	Create(ctx context.Context, path string, signature *regula.Signature) error
+	// Put is used to create a new version of the ruleset.
```
YassineE:
```suggestion
	// Put is used to add rules to a ruleset. It creates a new version of the ruleset.
```

yaziine (approved):
Noice 👍

rule/cast_expr.go
```diff
@@ -50,7 +46,7 @@ func (n *exprIntToFloat) Eval(params Params) (*Value, error) {
 	if err != nil {
 		return nil, err
 	}
-	iVal, err := strconv.ParseInt(val.Data, 10, 64)
+	iVal, err := val.ToInt64()
```
tealeg: :heart_eyes:

rule/cast_expr.go
```diff
@@ -99,7 +95,7 @@ func (n *exprFloatToInt) Eval(params Params) (*Value, error) {
 		return nil, err
 	}
 	// Handily this will just give us the int we want ;-)
-	fVal, err := strconv.ParseFloat(val.Data, 64)
+	fVal, err := val.ToFloat64()
```
tealeg: :hugs:

ruleset.go
```diff
@@ -123,19 +80,25 @@ func (r *Ruleset) validate() error {
 
 // Signature represents the signature of a ruleset.
 type Signature struct {
-	ReturnType string
-	ParamTypes map[string]string
+	ReturnType string            `json:"returnType"`
+	Params     map[string]string `json:"params"` // TODO(asdine) rename to Params
```
tealeg: The TODO troubles me - rename what to Params?

store/etcd/put.go
```diff
@@ -17,222 +14,146 @@ import (
...
+	return ruleset, err
+}
 
-		err = s.putVersions(stm, path, version)
-		if err != nil {
-			return err
-		}
+// rulesPutter is responsible for validating and storing rules, updating checksums and other actions
+// that are required in order to add a new ruleset version correctly.
+type rulesPutter struct {
```
tealeg: This "Putter" wording amuses me every time I see it.  "Putter" is a real word in English, but it doesn't describe something that "puts".  Here are the definitions:
```
 putter (pŭtˈər)►

    n.
    A short golf club used for putting.
    n.
    A golfer who is putting.
    v.
    To occupy oneself in an aimless or ineffective manner.
```
Funnily enough, I don't think there is a noun for someone or something that "puts", so I've no better suggestion.  I'll just carry on being amused.

tealeg (approved):
Blimey.  What a magnificent beast of a PR.

Clean as a whistle as ever.  Lots of love for type conversions built into `Value` expressions.

ruleset.go
asdine: I forgot to remove the TODO :laughing:

store/etcd/put.go
asdine: It troubled me too! To be honest I wrote it expecting that you'd propose a better one if this one was too weird. But you failed. Shame on you.

PR #95, opened 2019-03-14
Refactor packages

asdine:
This PR changes the naming of certain packages:

- `api`, `api/client`, `api/server` -> `http`, `http/client`, `http/server`

- `store`, `store/etcd` -> `api`, `api/etcd`

The `api` package (the new one) defines how the Regula API works and provides an implementation (the etcd version). All the API logic is contained within this package which allows the `http` layer to be very thin and dumb. The name `store` was a little bit too database centric, whilst it contained much more logic than just storage.


YassineE:
👍

api/etcd/rulesets_test.go
```diff
@@ -648,18 +648,26 @@ func TestWatch(t *testing.T) {
...
 	require.Len(t, events.Events, 2)
 	require.NotEmpty(t, events.Revision)
-	require.Equal(t, store.RulesetPutEvent, events.Events[0].Type)
+	require.Equal(t, api.RulesetPutEvent, events.Events[0].Type)
 	require.Equal(t, "ab", events.Events[0].Path)
-	require.Equal(t, store.RulesetPutEvent, events.Events[1].Type)
+	require.Equal(t, api.RulesetPutEvent, events.Events[1].Type)
 	require.Equal(t, "a/1", events.Events[1].Path)
+
+	t.Run("timeout", func(t *testing.T) {
```
tealeg: :tulip: generally its better not to do stuff like this when you're doing a large, but otherwise fairly mechanical change.   Not that there's anything wrong with the code, it's just that I've found people (naturally) tend to be less scrupulous about reviews where (for example) we're just renaming things, and more substantive changes in their midst could be lost.

asdine: You're absolutely right, next time I'll simply submit two different PRs

PR #98, opened 2019-03-29
Add latest ruleset page

### LINKED ISSUE #32: Consult latest ruleset page
Create a page that allows consulting the latest ruleset version

Path: `GET /rulesets/<path>/latest`
Get data from: `GET /i/rulesets/<path>`

Sketch: 
regula-040-ruleset-latest

Related to #33

YassineE:
There is something that bothers me, now the "/" from the URL are encoded because of the separator in the path. 
For the given ruleset path: `boost/hidden/v2`, the vuejs's router think that it's an endpoint instead of a path in the router declaration thus the `/rulesets/:path/latest` won't be reached.

Can we do something to avoid that?

This PR fixes #32.

### DIFF
--- a/ui/app/src/views/LatestRuleset/LatestRuleset.vue
+++ b/ui/app/src/views/LatestRuleset/LatestRuleset.vue
@@ -1,4 +1,134 @@
 <template>
-  <div>
-  </div>
+  <v-container id="consult">
+    <h1>{{path}}</h1>
+
+    <v-layout>
+      <v-flex md2 mt-5>
+        <v-toolbar color="grey" dark>
+          <v-toolbar-title>Parameters</v-toolbar-title>
+        </v-toolbar>
+        <v-card class="height-card scroll">
+          <v-card-text v-for="param in ruleset.signature.params" :key="param.name">{{param.name}}: {{param.type}}</v-card-text>
+        </v-card>
+      </v-flex>
+
+      <v-flex md2 ma-5>
+        <v-toolbar color="grey" dark>
+          <v-toolbar-title>Return type</v-toolbar-title>
+        </v-toolbar>
+        <v-card class="height-card">
+          <v-card-text>{{ruleset.signature.returnType}}</v-card-text>
+        </v-card>
+      </v-flex>
+
+      <v-flex md2 ma-5>
+        <v-toolbar color="grey" dark>
+          <v-toolbar-title>Versions</v-toolbar-title>
+        </v-toolbar>
+        <v-card class="height-card scroll">
+          <v-card-text v-for="version in ruleset.versions" :key="version">{{version}}</v-card-text>
+        </v-card>
+      </v-flex>
+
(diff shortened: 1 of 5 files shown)


ui/app/src/views/LatestRuleset/LatestRuleset.vue
```diff
@@ -1,4 +1,134 @@
...
+          <v-card-text v-for="version in ruleset.versions" :key="version">{{version}}</v-card-text>
+        </v-card>
+      </v-flex>
+
+      <v-flex md3></v-flex>
+
+      <v-flex md3 mt-5>
+        <!-- change the link to edit page -->
+        <router-link to="/rulesets/new">
+          <v-btn
```
tealeg: can this not be wrapped up into one line?

ui/app/src/views/LatestRuleset/LatestRuleset.vue
```diff
@@ -1,4 +1,134 @@
...
+  },
+
+  props: {
+    path: {
+      type: String,
+    }
+  },
+
+  data() {
+    return {
```
tealeg: Because JS, and particularly Vue, is not our "normal" work, I think we should try to be especially diligent about adding comments / docstrings to methods.  What seems obvious today, might be obscure in a months time.

ui/app/src/views/LatestRuleset/LatestRuleset.vue
```diff
@@ -1,4 +1,134 @@
...
+    }
+  },
+
+  mounted() {
+    this.fetchRuleset();
+  },
+
+  methods: {
+    fetchRuleset() {
+      axios
```
asdine: No need to make a call to the backend if you are going to ignore it, just fill the vue data directly

ui/app/src/views/LatestRuleset/LatestRuleset.vue
YassineE: Sure.

ui/app/src/views/LatestRuleset/LatestRuleset.vue
YassineE: I agree but for this one I think that it's not necessary to add comment. 
The `data` property is the main important things to know in Vue, it's for that reason that I didn't add comment, for example, it's like adding comment on top of the `main` function in a Go program.

YassineE:
>Also, the rules are still editable. Even though it doesn't change data on the server, it's poor UX. We should only display the rules without being able to change them.

Are you sure? Because in my machine the rules aren't editable thanks to the `editMode` props.

Thank you for the plugin suggestion, I installed them, but the errors aren't fixed on saving.

YassineE:
@asdine 🆙 
I fixed the eslint errors.
Can you make sure of this:
> Also, the rules are still editable. Even though it doesn't change data on the server, it's poor UX. We should only display the rules without being able to change them.

Issue #101, opened 2019-04-01
Add Gt, Gte, Lt, Lte operators


asdine:
Same behaviour as https://github.com/heetch/regula/blob/b79b8edceafa6f3c5d7461ba29f83cbbb94ddb40/rule/eq_expr.go#L8-L14

PR #103, opened 2019-04-02
Add Gt operator

YassineE:
Add Greater Than operator (partial fix of #101).

### DIFF
--- a/rule/expr.go
+++ b/rule/expr.go
@@ -241,10 +241,10 @@ type exprGT struct {
 	operator
 }
 
-// Gt creates an expression that takes at least two operands and
+// GT creates an expression that takes at least two operands and
 // evaluates to true if each successive operand has a higher value than
 // the next.
-func Gt(v1, v2 Expr, vN ...Expr) Expr {
+func GT(v1, v2 Expr, vN ...Expr) Expr {
 	return &exprGT{
 		operator: operator{
 			kind:     "gt",
@@ -255,7 +255,7 @@ func Gt(v1, v2 Expr, vN ...Expr) Expr {
 
 func (n *exprGT) Eval(params Params) (*Value, error) {
 	if len(n.operands) < 2 {
-		return nil, errors.New("invalid number of operands in Gt func")
+		return nil, errors.New("invalid number of operands in GT func")
 	}
 
 	vA, err := n.operands[0].Eval(params)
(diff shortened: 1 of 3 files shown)


rule/expr.go
```diff
@@ -237,6 +237,46 @@ func (n *exprIn) Eval(params Params) (*Value, error) {
 	return BoolValue(false), nil
 }
 
+type exprGt struct {
```
tealeg: I strongly suggest we use `GT` rather than `Gt` to remove the need to rename when when moving to 0.0.7.

YassineE: Done https://github.com/heetch/regula/pull/103/commits/35f736165459179b6bce32af3867a38ec6d8ec07

rule/expr.go
```diff
@@ -382,6 +422,45 @@ func (v *Value) Equal(other *Value) bool {
 	return v.compare(token.EQL, other)
 }
 
+// GT reports whether v is greater than other.
+func (v *Value) GT(other *Value) bool {
+	switch v.Type {
+	case "bool":
+		v1, _ := strconv.ParseBool(v.Data)
```
asdine: Why ignoring the error when you can return it?

YassineE: Because the function won't fail, I am sure that I will parse the correct type.
It's impossible to have a value in `v.Data` that mismatches the type in `v.Type`.

asdine: What if I create it by hand in the seeds scripts for example?

YassineE: The only way to create `Value`s is to use the `rule.TValue` functions (where `T` is a type) and those are strongly typed thus I don't see how we can have a value that differs from the type?

asdine: I'm just being picky here, but Regula is a lib and you don't know how people might want to use it:

```go
v := rule.Value{
  Type: "int64",
  Name: "foo",
  Data: fmt.Sprintf("%f", math.Pi / 2), // oups it's a float
}
```

YassineE: From that perspective, you are right. But for this "intermediate" version I think that we can ship it like that and be more carefully in the next one.

asdine: I don't think intermediate versions allow us to be less careful about what we are producing. I agree that we should not make too much effort for nothing, but here we are talking about adding an `if err != nil` that doesn't cost much.

YassineE: > Regula is a lib and you don't know how people might want to use it

This argument is enough to add the check. From time to time I forgot that the users can do whatever they want with the exported type and my opinion is based on what they are doing and how they are using it at the moment.
Will add the check in #107.

PR #111, opened 2019-05-10
Edit ruleset rules endpoint

### LINKED ISSUE #36: Edit ruleset rules endpoint
Endpoint: `PATCH /i/rulesets/<path>`

Expected payload :
```js
{
  "rules": [
        {
            "sExpr": "",
            "returnValue": "",
        }
    ]
}
```

The server must replace all the rules of the ruleset by the ones received in the request.

## Success
Response: 200

## Failure
Requirements: The `rules` field of the payload must contain at least one rule

Validation error:
```js
{ 
	"error": "validation",
	"fields": [
		{
			"path": ["rules", "2", "sExpr"],
			"error": {
				"message": "",
				"line": 10,
				"char": 11,
				"absChar": 120,
			}
		},
		{
			"path": ["rules", "2", "returnValue"],
			"error": {
				"message": "",
			}
		}
	]
}
```

tealeg:
Fixes #36 

This PR adds a handler for editing a Rulesets rules.  

Along the way I also:
- corrected the `PutFn` function of the mock `RulesetService`
- added a convenience method for deriving `sexpr.Paramaters` from `regula.Signature` types

### DIFF
--- a/ui/handler.go
+++ b/ui/handler.go
@@ -121,6 +121,32 @@ func (h *internalHandler) handleNewRulesetRequest(w http.ResponseWriter, r *http
 
 }
 
+func (h *internalHandler) handleEditRulesetRequest(w http.ResponseWriter, r *http.Request) {
+	path := strings.TrimPrefix(r.URL.Path, "/rulesets/")
+	if path == "" {
+		writeError(w, r, nil, http.StatusNotFound)
+		return
+	}
+
+	entry, err := h.service.Get(r.Context(), path, "")
+	if err != nil {
+		if err == store.ErrNotFound {
+			writeError(w, r, err, http.StatusNotFound)
+			return
+		}
+
+		writeError(w, r, err, http.StatusInternalServerError)
+		return
+	}
+	if entry == nil {
+		err := fmt.Errorf("No Ruleset found at path: %q", path)
+		writeError(w, r, err, http.StatusNotFound)
+		return
+	}
+	fmt.Printf("%v", entry)
+	reghttp.EncodeJSON(w, r, nil, http.StatusNoContent)
+}
+
 func (h *internalHandler) handleSingleRuleset(w http.ResponseWriter, r *http.Request) {
 	path := strings.TrimPrefix(r.URL.Path, "/rulesets/")
 
@@ -216,6 +242,8 @@ func (h *internalHandler) handleListRequest(w http.ResponseWriter, r *http.Reque
 func (h *internalHandler) rulesetsHandler() http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.Method {
+		case "PATCH":
+			h.handleEditRulesetRequest(w, r)
 		case "POST":
 			h.handleNewRulesetRequest(w, r)
 		case "GET":
--- a/ui/handler.go
+++ b/ui/handler.go
@@ -149,9 +149,9 @@ func (h *internalHandler) handleEditRulesetRequest(w http.ResponseWriter, r *htt
 	}
 
 	// Update the entry with the new rules
-	err = updateEntry(entry, nrr)
+	err, status := updateEntry(entry, nrr)
 	if err != nil {
-		writeError(w, r, err, http.StatusInternalServerError)
+		writeError(w, r, err, status)
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


ui/handler.go
```diff
@@ -121,6 +121,94 @@ func (h *internalHandler) handleNewRulesetRequest(w http.ResponseWriter, r *http
...
+	if path == "" {
+		writeError(w, r, nil, http.StatusNotFound)
+		return
+	}
+
+	nrr := &newRulesetRequest{}
+
+	err := json.NewDecoder(r.Body).Decode(nrr)
+	if err != nil {
+		writeError(w, r, err, http.StatusInternalServerError)
```
asdine: This should be a `http.StatusBadRequest` instead as it's the client's fault

ui/handler.go
```diff
@@ -121,6 +121,94 @@ func (h *internalHandler) handleNewRulesetRequest(w http.ResponseWriter, r *http
...
+		if err == store.ErrNotFound {
+			writeError(w, r, err, http.StatusNotFound)
+			return
+		}
+
+		writeError(w, r, err, http.StatusInternalServerError)
+		return
+	}
+	if result == nil {
+		err := fmt.Errorf("No Ruleset found at path: %q", path)
```
asdine: :tulip: No uppercase on errors

ui/handler.go
```diff
@@ -121,6 +121,94 @@ func (h *internalHandler) handleNewRulesetRequest(w http.ResponseWriter, r *http
...
+		writeError(w, r, err, http.StatusNotFound)
+		return
+	}
+	reghttp.EncodeJSON(w, r, nil, http.StatusNoContent)
```
asdine: Do we need to call `EncodeJSON` for that?

Does that work too?

```go
w.WriteHeader(http.StatusNoContent)
```

ui/handler.go
```diff
@@ -121,6 +121,94 @@ func (h *internalHandler) handleNewRulesetRequest(w http.ResponseWriter, r *http
...
+	err, status := updateEntry(entry, nrr)
+	if err != nil {
+		writeError(w, r, err, status)
+		return
+	}
+
+	// Write the new entry back to the DB
+	result, err := h.service.Put(r.Context(), path, entry.Ruleset)
+	if err != nil {
+		if err == store.ErrNotFound {
```
YassineE: The `Put` method will never return the `store.ErrNotFound` error. 
Only `store.ErrNotModified` can be returned.

ui/handler.go
```diff
@@ -121,6 +121,94 @@ func (h *internalHandler) handleNewRulesetRequest(w http.ResponseWriter, r *http
...
+	if result == nil {
+		err := fmt.Errorf("No Ruleset found at path: %q", path)
+		writeError(w, r, err, http.StatusNotFound)
+		return
+	}
+	reghttp.EncodeJSON(w, r, nil, http.StatusNoContent)
+}
+
+// updateEntry augments an existing entry with new Rules
+func updateEntry(entry *store.RulesetEntry, nrr *newRulesetRequest) (error, int) {
```
YassineE: Is not a Go antipattern to return the error first?

ui/handler.go
```diff
@@ -134,6 +222,8 @@ func (h *internalHandler) handleSingleRuleset(w http.ResponseWriter, r *http.Req
 
 	entry, err := h.service.Get(r.Context(), path, "")
 	if err != nil {
+		logger := reghttp.LoggerFromRequest(r)
+		logger.Debug().Msg("foo")
```
YassineE: I think that this debug can be removed.

ui/handler.go
YassineE: When saying "Only", I mean only ErrNotModified can be returned from our internal types error.

ui/handler.go
tealeg: Good point - I went this way because it's additional data that you only care about when there is an error, but it definitely looks weird.

ui/handler.go
tealeg: Agreed :-)

ui/handler.go
tealeg: Done

ui/handler.go
tealeg: Done

ui/handler.go
tealeg: :+1:

ui/handler.go
tealeg: You are correct.  Thanks, good catch.

ui/handler.go
```diff
@@ -121,6 +121,89 @@ func (h *internalHandler) handleNewRulesetRequest(w http.ResponseWriter, r *http
...
+	status, err := updateEntry(entry, nrr)
+	if err != nil {
+		writeError(w, r, err, status)
+		return
+	}
+
+	// Write the new entry back to the DB
+	result, err := h.service.Put(r.Context(), path, entry.Ruleset)
+	if err != nil {
+		writeError(w, r, err, http.StatusInternalServerError)
```
YassineE: I am afraid that if a ruleset is submitted without any changes the `Put` method will return `store.ErrNotModified` and here we will return `http.StatusInternalServerError`.

tealeg: OK. I'll add a test case and handle that appropriately.  :+1:

tealeg: DOne

tealeg: :up: @YassineE

PR #118, opened 2019-05-21
API simplification

asdine:
Package `regula`: Removed redundant ruleset validation code, the `api` package already does that

Package `errors`: Removed unnecessary errors, these errors should not be "catchable" and represent something bad that shouldn't happen in normal conditions

Package `api`: 
• `List` now returns a simple list of paths, with pagination. No prefix.
• Put now returns the version string, not the ruleset
• Renamed `ContinueToken` to `Cursor`

Package `api/etcd`:
• The change of behaviour of the `List` and `Put` APIs simplified a lot the etcd implementation. Lots of code removed
• Tests have been split into multiple files

Package `http`:
• Tests have been rewritten into table tests and split
• Applied changes from the API

:warning: Tests are failing on purpose, next PRs will fix all of that


api/etcd/get.go
```diff
@@ -49,28 +49,34 @@ func (s *RulesetService) Get(ctx context.Context, path, version string) (*regula
...
 	}
 
-	// decode versions
-	var versions pb.Versions
-	err = proto.Unmarshal(resp.Responses[1].GetResponseRange().Kvs[0].Value, &versions)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to unmarshal versions")
+	// the versions must be parsed from the keys returned by the second operation
+	versions := make([]string, len(resp.Responses[1].GetResponseRange().Kvs))
+	for i, kv := range resp.Responses[1].GetResponseRange().Kvs {
```
tealeg: Why don't we create a variable:
```go
kvs := resp.Responses[1].GetResponseRange().Kvs
```
.. and then do:
```
        versions := make([]string, len(kvs)
	for i, kv := range kvs {
```

api/etcd/get.go
```diff
@@ -49,28 +49,34 @@ func (s *RulesetService) Get(ctx context.Context, path, version string) (*regula
...
+		}
+	} else if version != "" {
+		// if no rules are returned but a version is specified, it means that the version doesn't exist
+		return nil, api.ErrRulesetNotFound
 	}
+
+	// if the version wasn't specified, get the latest returned
 	if version == "" {
-		_, version = s.pathVersionFromKey(string(resp.Responses[2].GetResponseRange().Kvs[0].Key))
+		version = versions[len(versions)-1]
```
tealeg: We could also have stored the result of the `len` we did on the `kvs` above.  Micro-optimisation, mind you ;-)

api/etcd/get_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+		ruleset1, err := s.Get(context.Background(), path, "")
+		require.NoError(t, err)
+		require.Equal(t, path, ruleset1.Path)
+		require.Len(t, ruleset1.Versions, 1)
+		require.NotEmpty(t, rules1, ruleset1.Version)
+		require.Equal(t, rules1, ruleset1.Rules)
+		require.Equal(t, sig, ruleset1.Signature)
+
+		// we are waiting 1 second here to avoid creating the new version in the same second as the previous one
```
tealeg: :tulip: given the constraint, we should perhaps guard against creating more than one per second if we wish to maintain version ordering this way (KSUID is a good solution, I'm not questioning that).

api/etcd/list_test.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/heetch/regula/api"
+	"github.com/heetch/regula/rule"
+	"github.com/stretchr/testify/require"
+)
+
+// List returns all rulesets rulesets or not depending on the query string.
```
tealeg: I think you need to rewrite this docstring, I'm not really sure what its trying to say.

api/service.go
```diff
@@ -3,54 +3,58 @@ package api
...
-	AllVersions   bool // return all versions of each rulesets
+	Limit  int    // If the Limit is lower or equal to 0 or greater than 100, it will be set to 50 by default.
+	Cursor string // pagination cursor
+}
+
+// GetLimit returns a limit that is withing the 0 - 100 boundries.
+// If the limit is incorrect, it returns 50.
+func (l *ListOptions) GetLimit() int {
+	if l.Limit <= 0 || l.Limit > 100 {
+		return 50 // TODO: make this one configurable
```
tealeg: Unless you plan to implement this change in an immediately following PR, please transfer it to an issue.  TODO's are where good ideas go to die ;-)

api/etcd/list.go
```diff
@@ -4,200 +4,61 @@ import (
...
 	var key string
 
-	limit := computeLimit(opt.Limit)
+	// only fetch keys
+	opts = append(opts, clientv3.WithKeysOnly())
 
-	if opt.ContinueToken != "" {
-		lastPath, err := base64.URLEncoding.DecodeString(opt.ContinueToken)
+	// if a cursor is specified, decode the key from it and start the request from that key
+	if opt.Cursor != "" {
```
YassineE: Nice wording 👍

api/etcd/list.go
```diff
@@ -4,200 +4,61 @@ import (
...
-	}
-}
+	// limit the number of results
+	opts = append(opts, clientv3.WithLimit(int64(opt.GetLimit())))
 
-// listPathsOnly returns only the path for each ruleset.
-func (s *RulesetService) listPathsOnly(ctx context.Context, key, prefix string, limit int, opts []clientv3.OpOption) (*api.Rulesets, error) {
-	opts = append(opts, clientv3.WithKeysOnly())
-	resp, err := s.Client.KV.Get(ctx, s.latestVersionPath(key), opts...)
+	// fetch signatures
```
YassineE: At first glance, it is not clear why you retrieve the signatures whereas the in-line doc says that only paths will be returned. I suggest to add more information here.
```suggestion
	// fetch signatures, the paths will be computed from signature keys
```

api/etcd/get_test.go
```diff
@@ -0,0 +1,69 @@
...
+		require.NoError(t, err)
+		require.Equal(t, path, ruleset2.Path)
+		require.Len(t, ruleset2.Versions, 2)
+		require.Equal(t, ruleset1.Version, ruleset2.Versions[0])
+		require.Equal(t, ruleset2.Version, ruleset2.Versions[1])
+		require.Equal(t, rules2, ruleset2.Rules)
+		require.Equal(t, sig, ruleset2.Signature)
+
+		// it should return the second version
+		rs, err := s.Get(context.Background(), path, ruleset2.Version)
```
YassineE: I suggest to ask for the 1st version because it is not the one returned by default. Wdyt?

api/etcd/get_test.go
asdine: Indeed, that's a very good point. I was thinking that in terms of UI, if two people update the same ruleset at the same moment, there's no real value in knowing who pressed the save button first. But if you use Regula's API to programmatically create rules sequentially you will lose the expected ordering.
I've created an issue for that: https://github.com/heetch/regula/issues/120

api/etcd/get.go
asdine: Actually, IIRC `len` is constant time, unlike C's `strlen` which is O(n) and justifies caching the length to reuse it.

api/service.go
asdine: Done :+1:  https://github.com/heetch/regula/issues/121

asdine:
:up: @tealeg

api/etcd/get.go
```diff
@@ -11,23 +11,23 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Get returns the ruleset related to the given path. By default, it returns the latest one.
-// It returns the related ruleset version if it's specified.
+// Get returns the ruleset related to the given path.
```
rogpeppe: ISTM there's still some logic relating to the version that could usefully be documented here.
For example, if the version is empty, it returns the latest version.

api/service.go
```diff
@@ -3,54 +3,58 @@ package api
...
-// ListOptions contains list options.
-// If the Limit is lower or equal to 0 or greater than 100, it will be set to 50 by default.
+// ListOptions is used to customize the List output.
 type ListOptions struct {
-	Limit         int
-	ContinueToken string
-	PathsOnly     bool // return only the paths of the rulesets
-	AllVersions   bool // return all versions of each rulesets
+	Limit  int    // If the Limit is lower or equal to 0 or greater than 100, it will be set to 50 by default.
+	Cursor string // pagination cursor
```
rogpeppe: Perhaps mention that if this is empty, the list starts from the beginning?

api/service.go
```diff
@@ -3,54 +3,58 @@ package api
...
 
-// ListOptions contains list options.
-// If the Limit is lower or equal to 0 or greater than 100, it will be set to 50 by default.
+// ListOptions is used to customize the List output.
 type ListOptions struct {
-	Limit         int
-	ContinueToken string
-	PathsOnly     bool // return only the paths of the rulesets
-	AllVersions   bool // return all versions of each rulesets
+	Limit  int    // If the Limit is lower or equal to 0 or greater than 100, it will be set to 50 by default.
```
rogpeppe: // Limit holds the maximum number of items to return.
// It it's not positive or greater than 100, a limit of 50 will be used.

?
Although it seems slightly odd that I can explicitly specify a limit of 100 and I'll get 100 items, but if I specify 101, I'll only get 50. Why not cap it at 100?

api/etcd/list.go
```diff
@@ -4,200 +4,61 @@ import (
...
-
-	// if a prefix is provided it must always return results
-	// otherwise it doesn't exist.
-	if resp.Count == 0 && prefix != "" {
-		return nil, api.ErrRulesetNotFound
-	}
-
-	ops := make([]clientv3.Op, 0, resp.Count)
-	txn := s.Client.KV.Txn(ctx)
+	rulesets.Paths = make([]string, 0, len(resp.Kvs))
```
rogpeppe: Assuming resp.Kvs is a map, it might be worth considering sorting the final slice so that the results are deterministic?
:tulip: :tulip:

api/etcd/get.go
tealeg: :+1: I knew it was constant time, but pushing it to the register, I didn't.

api/service.go
asdine: Good idea :+1:

api/etcd/list.go
asdine: Actually, `resp.Kvs` is a slice so the order is kept (I know the name is not the best)

Issue #112, opened 2019-05-10
Scaling the caching system


asdine:
Regula is designed to be used used by any kind of program: mobile apps, microservices, monoliths, frontend applications, etc.
Programs can use Regula in two ways:

### Using the Eval API

The client runs a synchronous http call to the Regula server in order to evaluate a ruleset. It can pass parameters and specify which version of the ruleset to use. Basically, this can be summed up as the following curl command:

```sh
curl "http://localhost:5331/rulesets/some/path?eval&param1=a&param2=b"
```

### By embedding Regula

The program imports Regula as a library, loads a bunch of ruleset definitions from the server, caches that in memory and evaluates everything locally.

## The issue with cache prefilling

The second option is great to avoid network round trips but the "loads a bunch of ruleset definitions from the server" is problematic.

Currently, the client can load these rulesets by selecting a `prefix`:

_Every ruleset whose path begins with this prefix will get downloaded, including all the different versions of these rulesets._

Of course, an empty prefix is a valid one, meaning that it is possible to tell the client to download every possible ruleset including their precedent versions.

This obviously doesn't scale at all.

This could be solved by selecting a long prefix to narrow the results, but any node of a tree is a tree itself so it would only help providing that we are sure that node doesn't grow big.

## Let's review our requirements

Regula was designed with a set of requirements in mind and I believe that wanting to absolutely comply with all of them will lead to a bad design. Here is the list:

- Only one network round trip (to fill the local cache)
- Have access to any ruleset regardless of its path in the tree without any additional round trip
- Have access to any precedent version of any ruleset regardless of its path in the tree without any additional round trip
- Being notified if any new version is created so the program can update its cache

By brainstorming with @tealeg, it appeared to us that we focused too much on _making sure every ruleset is available locally just in case you might need it_, whereas in fact people actually know what rulesets they need within their program.

Here is a revised list of requirements that in our opinion scales much better, while providing a great experience:

- Only one network round trip (to fill the local cache) for _selected rulesets_
- Have access to any ruleset regardless of its path in the tree ~without any additional round trip~
- Have access to any precedent version of any ruleset regardless of its path in the tree ~without any additional round trip~
- Being notified if any new version is created so the program can update its cache

With these new requirements, we can provide the following solution.

## The solution

In order to provide a solution that works in multiple situations, I will first describe the default behavior then talk about customizations that will lead to satisfying the requirements listed above.

### Default client

[...]

PR #122, opened 2019-05-22
New watch api

asdine:
This PR refreshes the Watch API according to what has been decided in issue #112 

The Watch API is now accessible under the following endpoint:
```
// URL
POST /rulesets/?watch

// Optional query params
?revision

// Payload
{
  "paths": [
    "pathA",
    "pathB",
   ],
  "revision": 10
}
```

The body specifies which paths you want to watch.
An empty body means watch everything.

This PR also contains a few improvements in the `server` package

### DIFF
--- a/api/etcd/rulesets_test.go
+++ b/api/etcd/rulesets_test.go
@@ -25,12 +25,10 @@ var (
 	endpoints   = []string{"localhost:2379", "etcd:2379"}
 )
 
-func init() {
-	rand.Seed(time.Now().UnixNano())
-}
-
 func newEtcdRulesetService(t *testing.T) (*RulesetService, func()) {
 	t.Helper()
+
+	rand.Seed(time.Now().UnixNano())
 
 	cli, err := clientv3.New(clientv3.Config{
 		Endpoints:   endpoints,
--- a/api/etcd/rulesets_test.go
+++ b/api/etcd/rulesets_test.go
@@ -50,6 +50,8 @@ func newEtcdRulesetService(t *testing.T) (*RulesetService, func()) {
 }
 
 func createRuleset(t *testing.T, s *RulesetService, path string, rules ...*rule.Rule) *regula.Ruleset {
+	t.Helper()
+
 	_, err := s.Put(context.Background(), path, rules)
 	if err != nil && err != api.ErrRulesetNotModified {
 		require.NoError(t, err)
@@ -61,6 +63,8 @@ func createRuleset(t *testing.T, s *RulesetService, path string, rules ...*rule.
 }
 
 func createBoolRuleset(t *testing.T, s *RulesetService, path string, rules ...*rule.Rule) *regula.Ruleset {
+	t.Helper()
+
 	err := s.Create(context.Background(), path, &regula.Signature{ReturnType: "bool"})
 	assert.False(t, err != nil && err != api.ErrAlreadyExists)
 	return createRuleset(t, s, path, rules...)
--- a/api/etcd/rulesets_test.go
+++ b/api/etcd/rulesets_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/heetch/regula"
 	"github.com/heetch/regula/api"
 	"github.com/heetch/regula/rule"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -24,7 +25,7 @@ var (
 	endpoints   = []string{"localhost:2379", "etcd:2379"}
 )
 
-func Init() {
+func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
@@ -55,12 +56,12 @@ func createRuleset(t *testing.T, s *RulesetService, path string, rules ...*rule.
 	}
 
 	rs, err := s.Get(context.Background(), path, "")
-	require.NoError(t, err)
+	assert.NoError(t, err)
 	return rs
 }
 
 func createBoolRuleset(t *testing.T, s *RulesetService, path string, rules ...*rule.Rule) *regula.Ruleset {
 	err := s.Create(context.Background(), path, &regula.Signature{ReturnType: "bool"})
-	require.False(t, err != nil && err != api.ErrAlreadyExists)
+	assert.False(t, err != nil && err != api.ErrAlreadyExists)
 	return createRuleset(t, s, path, rules...)
 }
--- a/api/etcd/watch.go
+++ b/api/etcd/watch.go
@@ -43,7 +43,7 @@ func (s *RulesetService) Watch(ctx context.Context, paths []string, revision int
 			}
 
 			var list []api.RulesetEvent
-			for i, ev := range wresp.Events {
+			for _, ev := range wresp.Events {
 				// filter keys that haven't been selected
 				if !s.shouldIncludeEvent(ev, paths) {
 					s.Logger.Debug().Str("type", string(ev.Type)).Str("key", string(ev.Kv.Key)).Msg("watch: ignoring event key")
@@ -56,10 +56,6 @@ func (s *RulesetService) Watch(ctx context.Context, paths []string, revision int
 					continue
 				}
 
-				list = append(list, api.RulesetEvent{
-					Type: api.RulesetPutEvent,
-				})
-
 				var pbrs pb.Rules
 				err := proto.Unmarshal(ev.Kv.Value, &pbrs)
 				if err != nil {
@@ -67,9 +63,13 @@ func (s *RulesetService) Watch(ctx context.Context, paths []string, revision int
 					return nil, errors.Wrap(err, "failed to unmarshal entry")
 				}
 				path, version := s.pathVersionFromKey(string(ev.Kv.Key))
-				list[i].Path = path
-				list[i].Rules = rulesFromProtobuf(&pbrs)
-				list[i].Version = version
+
+				list = append(list, api.RulesetEvent{
+					Type:    api.RulesetPutEvent,
+					Path:    path,
+					Rules:   rulesFromProtobuf(&pbrs),
+					Version: version,
+				})
 			}
 
 			// none of the returned events matched the user selection
--- a/api/etcd/watch.go
+++ b/api/etcd/watch.go
@@ -12,13 +12,15 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Watch a list of paths for changes and return a list of events. If the list is empty or nil,
-// watch all paths.If the revision is negative, watch from the latest revision.
+// Watch a list of paths for changes and return a list of events. If paths is empty or nil,
+// watch all paths. If the revision is negative, watch from the latest revision.
 // This method blocks until there is a change if one of the paths or until the context is canceled.
 // The given context can be used to limit the watch period or to cancel any running one.
-func (s *RulesetService) Watch(ctx context.Context, paths []string, revision int64) (*api.RulesetEvents, error) {
+func (s *RulesetService) Watch(ctx context.Context, opt api.WatchOptions) (*api.RulesetEvents, error) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
+
+	revision := opt.Revision
 
 	opts := []clientv3.OpOption{clientv3.WithPrefix()}
 	if revision > 0 {
@@ -33,7 +35,7 @@ func (s *RulesetService) Watch(ctx context.Context, paths []string, revision int
 		select {
 		case wresp := <-wc:
 			if err := wresp.Err(); err != nil {
-				return nil, errors.Wrapf(err, "failed to watch paths: '%#v'", paths)
+				return nil, errors.Wrapf(err, "failed to watch paths: '%#v'", opt.Paths)
 			}
 
 			revision = wresp.Header.Revision
@@ -45,7 +47,7 @@ func (s *RulesetService) Watch(ctx context.Context, paths []string, revision int
 			var list []api.RulesetEvent
 			for _, ev := range wresp.Events {
 				// filter keys that haven't been selected
-				if !s.shouldIncludeEvent(ev, paths) {
+				if !s.shouldIncludeEvent(ev, opt.Paths) {
 					s.Logger.Debug().Str("type", string(ev.Type)).Str("key", string(ev.Kv.Key)).Msg("watch: ignoring event key")
 					continue
 				}
--- a/api/etcd/watch.go
+++ b/api/etcd/watch.go
@@ -15,6 +15,7 @@ import (
 
 // Watch a list of paths for changes and return a list of events. If the list is empty or nil,
 // watch all paths. If the revision is empty, watch from the latest revision.
+// This method blocks until there is a change if one of the paths or until the context is canceled.
 // The given context can be used to limit the watch period or to cancel any running one.
 func (s *RulesetService) Watch(ctx context.Context, paths []string, revision string) (*api.RulesetEvents, error) {
 	ctx, cancel := context.WithCancel(ctx)
--- a/api/etcd/watch.go
+++ b/api/etcd/watch.go
@@ -12,8 +12,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Watch the given prefix for anything new.
-func (s *RulesetService) Watch(ctx context.Context, prefix string, revision string) (*api.RulesetEvents, error) {
+// Watch a list of paths for changes and return a list of events. If the list is empty or nil,
+// watch all paths. If the revision is empty, watch from the latest revision.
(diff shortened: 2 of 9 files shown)


api/etcd/watch_test.go
```diff
@@ -8,55 +8,84 @@ import (
...
+			defer cancel()
 
-	wg.Wait()
+			var events api.RulesetEvents
+			var rev string
+			var watchCount int
+			for len(events.Events) != len(test.expected) && watchCount < 4 {
+				evs, err := s.Watch(ctx, test.paths, rev)
+				if err != nil {
+					assert.Equal(t, err, context.DeadlineExceeded)
```
rogpeppe: ISTM that if we've exceeded the deadline, we've got a test that takes 10 seconds, and even though that's the correct error, it might be better to fail the test with an different message.

```suggestion
if err != nil {
    if err == context.DeadlineExceeded {
       t.Errorf("timed out waiting for expected events")
    } else {
       t.Errorf("unexpected error from watcher: %v", err)
    }
   break
}
```

api/etcd/watch_test.go
```diff
@@ -8,55 +8,84 @@ import (
...
+					break
+				}
+				assert.True(t, len(evs.Events) >= 1)
```
rogpeppe:
```suggestion
assert.True(t, len(evs.Events) > 0)
```

http/server/api.go
```diff
@@ -18,28 +16,19 @@ import (
...
-	}
-
-	ctx, cancel := context.WithTimeout(r.Context(), s.timeout)
-	defer cancel()
-	r = r.WithContext(ctx)
+	path := r.URL.Path
 
 	switch r.Method {
 	case "GET":
 		if _, ok := r.URL.Query()["list"]; ok {
```
rogpeppe: Isn't it more conventional to use r.Form ? That way the values are parsed only once FWIW.

http/server/api.go
```diff
@@ -50,6 +39,15 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		s.get(w, r, path)
 		return
 	case "POST":
+		if _, ok := r.URL.Query()["watch"]; ok {
+			if len(path) != 0 {
```
rogpeppe: I wonder if it might be nicer to have an explicit path to use to watch (e.g. `/watch-rulesets`), rather than a POST to `/rulesets`. Not sure.
:tulip:

http/server/api.go
```diff
@@ -176,16 +174,32 @@ func (s *rulesetAPI) eval(w http.ResponseWriter, r *http.Request, path string) {
...
+// that happened since the start of the watch.
+// if the revision query param is specified, it returns anything that happened after that revision.
+// If no paths are specificied, it watches any path.
+// The request context can be used to limit the watch period or to cancel any running one.
+func (s *rulesetAPI) watch(w http.ResponseWriter, r *http.Request) {
+	var paths []string
+
+	// an empty body means watch all paths
+	if r.ContentLength > 0 {
+		err := json.NewDecoder(r.Body).Decode(&paths)
```
rogpeppe: As a rule, I think it's nicer to avoid top level arrays in POST bodies, as it makes it very hard to extend backward compatibly in future versions. How about using an object with a "paths" member instead? This also it makes it clearer to the client that they're specifying paths and not something else.

http/server/api_test.go
```diff
@@ -255,64 +257,51 @@ func TestServerWatch(t *testing.T) {
...
 
-		if test.status == http.StatusOK {
-			var res api.RulesetEvents
-			err := json.NewDecoder(w.Body).Decode(&res)
-			require.NoError(t, err)
-			if test.es != nil {
-				require.Equal(t, len(test.es.Events), len(res.Events))
-				for i := range l.Events {
-					require.Equal(t, l.Events[i], res.Events[i])
+			if test.status == http.StatusOK {
```
rogpeppe:
```suggestion
if test.status != http.StatusOK {
    return
}
```

to save a bit of indentation?

http/server/handler.go
```diff
@@ -37,21 +37,23 @@ func NewHandler(rsService api.RulesetService, cfg Config) http.Handler {
...
+		rulesets: rsService,
 	}
 
 	// router
 	mux := http.NewServeMux()
-	mux.HandleFunc("/rulesets/", func(w http.ResponseWriter, r *http.Request) {
+	mux.Handle("/rulesets/", http.StripPrefix("/rulesets/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if _, ok := r.URL.Query()["watch"]; ok && r.Method == "GET" {
-			rulesetsHandler.ServeHTTP(w, r.WithContext(cfg.WatchCancelCtx))
+			ctx, cancel := context.WithTimeout(cfg.WatchCancelCtx, cfg.WatchTimeout)
```
rogpeppe: Why does this need to second-guess the routing? Can't the watch code itself install its own timeout?

rogpeppe (changes requested):
This looks really nice in general, thanks. LGTM modulo a bunch of minor comments and suggestions, and one likely bug...

api/etcd/rulesets_test.go
```diff
@@ -24,7 +25,7 @@ var (
 	endpoints   = []string{"localhost:2379", "etcd:2379"}
 )
 
-func Init() {
+func init() {
```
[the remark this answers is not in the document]
asdine: Do you mean is it worth having an init method? or the `rand.Seed` call?

api/etcd/watch.go
```diff
@@ -12,43 +13,65 @@ import (
...
+// Watch a list of paths for changes and return a list of events. If the list is empty or nil,
+// watch all paths. If the revision is empty, watch from the latest revision.
+// The given context can be used to limit the watch period or to cancel any running one.
+func (s *RulesetService) Watch(ctx context.Context, paths []string, revision string) (*api.RulesetEvents, error) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
 	opts := []clientv3.OpOption{clientv3.WithPrefix()}
-	if i, _ := strconv.ParseInt(revision, 10, 64); i > 0 {
+	rev, _ := strconv.ParseInt(revision, 10, 64)
```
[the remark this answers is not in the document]
asdine: Good point, I'll change that to an int64

api/etcd/watch.go
```diff
@@ -12,43 +13,65 @@ import (
...
+		opts = append(opts, clientv3.WithRev(rev+1))
 	}
 
-	events := api.RulesetEvents{
-		Revision: revision,
-	}
+	var events api.RulesetEvents
 
-	wc := s.Client.Watch(ctx, s.rulesPath(prefix, ""), opts...)
+	wc := s.Client.Watch(ctx, s.rulesPath("", ""), opts...)
```
YassineE: Instead of watching the whole tree, I found out the [WithFromKey](https://godoc.org/github.com/coreos/etcd/clientv3#WithFromKey) option  maybe you can use it to specify the keys to watch?

YassineE: If it's doable it will reduce the complexity of this function a lot.

api/etcd/rulesets_test.go
```diff
@@ -55,12 +56,12 @@ func createRuleset(t *testing.T, s *RulesetService, path string, rules ...*rule.
...
 	rs, err := s.Get(context.Background(), path, "")
-	require.NoError(t, err)
+	assert.NoError(t, err)
 	return rs
 }
 
 func createBoolRuleset(t *testing.T, s *RulesetService, path string, rules ...*rule.Rule) *regula.Ruleset {
 	err := s.Create(context.Background(), path, &regula.Signature{ReturnType: "bool"})
-	require.False(t, err != nil && err != api.ErrAlreadyExists)
+	assert.False(t, err != nil && err != api.ErrAlreadyExists)
```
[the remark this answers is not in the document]
asdine: This function is a helper which means that it can be called anywhere, including in a goroutine.
Panic'ing in a goroutine (that's what require does) doesn't let you run the `defer` functions that are not in the same goroutine which can cause problems. Also, error messages when test fail are obscure. With `assert`, it's all fine, so I'm more for using assert in helpers and require in tests, though that's not a hard rule.

api/etcd/watch.go
asdine: Unfortunately, it only works for one key

api/etcd/rulesets_test.go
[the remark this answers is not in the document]
rogpeppe: the rand.Seed call.

api/etcd/rulesets_test.go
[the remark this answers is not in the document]
YassineE: Is it worth adding `t.Helper()` call in it?

api/etcd/watch.go
YassineE: Can't you chain the call?

api/etcd/watch.go
```diff
@@ -12,43 +13,65 @@ import (
...
+				// filter event types, keep only PUT events
+				if ev.Type != mvccpb.PUT {
 					s.Logger.Debug().Str("type", string(ev.Type)).Msg("watch: ignoring event type")
 					continue
 				}
 
+				list = append(list, api.RulesetEvent{
+					Type: api.RulesetPutEvent,
+				})
+
```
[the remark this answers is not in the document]
asdine: Absolutely, for both points :+1:

api/etcd/rulesets_test.go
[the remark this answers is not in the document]
asdine: Yes because in case a test fails and doesn't properly clean up the database, we might open the same namespace and have duplicate errors. That's actually what was happening  before I realized `init` was spelled wrongly

api/etcd/rulesets_test.go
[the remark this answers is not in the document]
asdine: Good idea, I'll add it

api/etcd/watch.go
asdine: Not for watches

YassineE: I mean `s.Client.Watch(ctx, s.rulespath(...), WithFromKey(k1), WithFromKey(k2),...)`.

api/etcd/watch_test.go
```diff
@@ -8,55 +8,84 @@ import (
...
-		defer wg.Done()
+			var wg sync.WaitGroup
 
-		time.Sleep(time.Second)
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
 
-		r := rule.New(rule.True(), rule.BoolValue(true))
+				time.Sleep(time.Second)
```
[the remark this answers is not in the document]
asdine: I don't like it either, the issue here is that I want to make sure I had the time to run the watch method in the other goroutine before writing data to the database.
I'll add a comment but I'm open to any suggestion

api/etcd/watch.go
asdine: `WithFromKey` doesn't take any argument, it uses the key (the second argument of `Watch`) as the key to watch

http/server/api.go
asdine: Actually, we need a POST anyway because we need to send a body in the request

api/etcd/rulesets_test.go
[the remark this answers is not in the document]
rogpeppe: oooh that's bad, and really should be dealt with by a helper IMHO. I'll just mention github.com/juju/postgrestest one more time :) :)

api/etcd/rulesets_test.go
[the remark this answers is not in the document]
rogpeppe: I have to say I think I agree with @YassineE here. If a testing helper takes a `*testing.T` argument and doesn't return an indication of success, I think it's reasonable to have it fail immediately, so the test doesn't proceed with partially wrong/confusing data. Perhaps a name like `requireCreateBoolRuleset` might provide more indication that it fails immediately on error?

api/etcd/rulesets_test.go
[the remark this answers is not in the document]
asdine: I've moved it to the test helper

asdine:
:up:

yaziine (approved):
👍

tealeg (approved):
Couple of minor things.

api/etcd/watch_test.go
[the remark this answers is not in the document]
rogpeppe: Do you actually need the watch statement to execute before this code? How about adding some rulesets, wait for an event to arrive, then adding some more, so you check both the initial case and the later case? (I'm not entirely sure of the intended semantics when there are rulesets already there though - what's meant to happen?)

api/service.go
```diff
@@ -55,10 +56,21 @@ func (l *ListOptions) GetLimit() int {
 	return l.Limit
 }
 
+// WatchOptions gives indications on what rulesets to watch.
+type WatchOptions struct {
+	// List of paths to watch for changes.
+	// If the slice is empty, watch all paths.
+	Paths []string
+	// Indicates from which revision start watching.
```
rogpeppe:
```suggestion
// Indicates from which revision to start watching.
```

http/server/api.go
rogpeppe: It was really the re-purposing of the endpoint for both purposes that felt odd to me. I don't mind much - we should really just be using gRPC and then we wouldn't need to care about such trivialities :)

http/server/api_test.go
```diff
@@ -119,11 +120,12 @@ func TestServerList(t *testing.T) {
...
 	}{
 		{"OK", "/rulesets/?list", http.StatusOK, &rss, api.ListOptions{}, nil},
-		{"WithLimitAndCursor", "/rulesets/a?list&limit=10&cursor=abc123", http.StatusOK, &rss, api.ListOptions{Limit: 10, Cursor: "abc123"}, nil},
+		{"WithLimitAndCursor", "/rulesets/?list&limit=10&cursor=abc123", http.StatusOK, &rss, api.ListOptions{Limit: 10, Cursor: "abc123"}, nil},
 		{"NoResult", "/rulesets/?list", http.StatusOK, new(api.Rulesets), api.ListOptions{}, nil},
-		{"InvalidCursor", "/rulesets/someprefix?list&cursor=abc123", http.StatusBadRequest, new(api.Rulesets), api.ListOptions{Cursor: "abc123"}, api.ErrInvalidCursor},
-		{"UnexpectedError", "/rulesets/someprefix?list", http.StatusInternalServerError, new(api.Rulesets), api.ListOptions{}, errors.New("unexpected error")},
-		{"InvalidLimit", "/rulesets/someprefix?list&limit=badlimit", http.StatusBadRequest, nil, api.ListOptions{}, nil},
+		{"InvalidCursor", "/rulesets/?list&cursor=abc123", http.StatusBadRequest, new(api.Rulesets), api.ListOptions{Cursor: "abc123"}, api.ErrInvalidCursor},
+		{"UnexpectedError", "/rulesets/?list", http.StatusInternalServerError, new(api.Rulesets), api.ListOptions{}, errors.New("unexpected error")},
```
rogpeppe: Why is this an internal server error rather than a bad request?

rogpeppe (approved):
Looks great in general, with one simplification suggestion, and perhaps it's worth improving that Watch test?

api/etcd/watch.go
```diff
@@ -12,61 +12,103 @@ import (
...
+			// caller can start after the filtered events.
+			events.Revision = revision
 			return &events, ctx.Err()
 		}
 	}
 }
+
+// shouldIncludeEvent reports whether the given event should be included
+// in the Watch data for the given paths.
+func (s *RulesetService) shouldIncludeEvent(ev *clientv3.Event, paths []string) bool {
```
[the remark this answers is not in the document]
asdine: `Put` is the only function responsible for writing in that path, I'd say that as long as `Put` is tested it should be fine, but I'm open to suggestions

api/etcd/watch_test.go
```diff
@@ -8,55 +8,93 @@ import (
...
-	require.Equal(t, "aa", events.Events[0].Path)
-	require.Equal(t, api.RulesetPutEvent, events.Events[0].Type)
+			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+			defer cancel()
 
-	wg.Wait()
+			var events api.RulesetEvents
+			var rev int64
+			var watchCount int
+			for len(events.Events) != len(test.expected) && watchCount < 4 {
```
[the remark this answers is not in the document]
asdine: It can't produce more events because each call to `createBoolRuleset` in the goroutine above will produce exactly one event

http/server/api_test.go
asdine: The 500 will not be caused by the URL by the error `errors.New("unexpected error")` that will be returned by the mock.
Basically, it's to simulate an unexpected error (network, etc.) happening in the store.

PR #123, opened 2019-05-22
Fix Create API

asdine:
The HTTP Create API was defined like this:

```
POST /rulesets/<path>
{// signature
  "returnType": "foo",
  "params": {}
} 
```

I changed it to
```
POST /rulesets/
{
  "path": "bar",
  "returnType": "foo",
  "params": {}
} 
```

which seems more correct

### DIFF
--- a/http/server/api.go
+++ b/http/server/api.go
@@ -58,7 +58,11 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		s.create(w, r, path)
+		if len(path) != 0 {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		s.create(w, r)
 		return
 	case "PUT":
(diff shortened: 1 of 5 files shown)


http/server/api.go
```diff
@@ -48,7 +48,11 @@ func (s *rulesetAPI) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		s.create(w, r, path)
+		if len(path) != 0 {
+			w.WriteHeader(http.StatusNotFound)
```
YassineE: Isn't returning http.StatusBadRequest more relevant here?

asdine: No because that's the wrong path: `/rulesets/p/a/t/h` and `/rulesets/` are different paths

yaziine (approved):

tealeg (approved):
:+1:

PR #125, opened 2019-05-29
Don't retry when client suffers http timeout

### LINKED ISSUE #19: Context timeout error should not be retriable
Currently, when a timeout occurs when using the API client, the request is retried 3 times with a canceled context. The request must not retry and return directly instead
https://github.com/heetch/regula/blob/edb915624cdc320986dc0416505b0083b20f01f5/api/client/client.go#L169-L171

tealeg:
Fixes #19 

This PR makes `url.Error` instances where `err.Timeout() == true` be excluded from the definition of `isRetriableError`, and implements a test case for it.

### DIFF
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -56,7 +56,7 @@ func (te *timeoutErr) Timeout() bool {
 
 // Error makes timeoutErr comply with the error interface.
 func (te *timeoutErr) Error() string {
-	return "net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers"
+	return "net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"
 }
 
 func TestTimeout(t *testing.T) {
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -77,7 +77,7 @@ func TestTimeout(t *testing.T) {
 	require.NoError(t, err)
 	_, err = c.tryN(context.Background(), req, nil, 3)
 	// // This *must* error
-	require.EqualError(t, err, "foo http://www.example.com/api/test: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers")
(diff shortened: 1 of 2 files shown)


api/client/client_test.go
```diff
@@ -27,3 +29,55 @@ func TestNewRequestPrefersSpecifiedUserAgentHeaderToDefault(t *testing.T) {
...
+type timeoutErr struct{}
+
+// Timeout makes timeoutErr comply with the private interface url.timeout
+func (te *timeoutErr) Timeout() bool {
+	return true
+}
+
+// Error makes timeoutErr comply with the error interface.
+func (te *timeoutErr) Error() string {
+	return "net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers"
```
YassineE:
```suggestion
	return "net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"
```

api/client/client_test.go
```diff
@@ -27,3 +29,55 @@ func TestNewRequestPrefersSpecifiedUserAgentHeaderToDefault(t *testing.T) {
...
+	impatientClient := HTTPClient(
+		&http.Client{
+			Transport: rtc})
+	c, err := New("http://www.example.com", impatientClient)
+	require.NoError(t, err)
+	req, err := c.newRequest("foo", "/api/test", nil)
+	require.NoError(t, err)
+	_, err = c.tryN(context.Background(), req, nil, 3)
+	// // This *must* error
+	require.EqualError(t, err, "foo http://www.example.com/api/test: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers")
```
YassineE:
```suggestion
	require.EqualError(t, err, "foo http://www.example.com/api/test: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)")
```

